PR #4, opened 2019-09-16
Added error handling to the dataset api response closing

dandwal:
### What

- Added error handling to the dataset api response closing
- Reordered download service auth token expected parameter position

### How to review

Check this is consistent with the other clients

### Who can review

Anyone except me

### DIFF
--- a/dataset/dataset.go
+++ b/dataset/dataset.go
@@ -101,7 +101,7 @@ func (c *Client) Get(ctx context.Context, id, serviceToken string) (m Model, err
 		return
 	}
 	defer closeResponseBody(ctx, resp)
-	
+
 	if resp.StatusCode != http.StatusOK {
 		err = NewDatasetAPIResponse(resp, uri)
 		return
@@ -151,7 +151,7 @@ func (c *Client) GetByPath(ctx context.Context, serviceToken, path string) (m Mo
 		return
 	}
 	defer closeResponseBody(ctx, resp)
-	
+
 	if resp.StatusCode != http.StatusOK {
 		err = NewDatasetAPIResponse(resp, uri)
 		return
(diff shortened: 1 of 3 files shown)


dataset/dataset.go
```diff
@@ -48,6 +49,20 @@ type Client struct {
 	url string
 }
 
+// closeResponseBodyWithContext closes the response body and logs an error containing the context if unsuccessful
+func closeResponseBodyWithContext(ctx context.Context, resp *http.Response) {
+	if err := resp.Body.Close(); err != nil {
+		log.ErrorCtx(ctx, err, log.Data{"Message": "error closing http response body"})
+	}
+}
+// closeResponseBody closes the response body and logs an error if unsuccessful
```
eldeal: comment is on the wrong line/dont need the empty line after it

dataset/dataset.go
```diff
@@ -48,6 +49,20 @@ type Client struct {
 	url string
 }
 
+// closeResponseBodyWithContext closes the response body and logs an error containing the context if unsuccessful
+func closeResponseBodyWithContext(ctx context.Context, resp *http.Response) {
+	if err := resp.Body.Close(); err != nil {
+		log.ErrorCtx(ctx, err, log.Data{"Message": "error closing http response body"})
```
eldeal: I believe logging standards say the keys in log.Data should be lower cased. Not sure whether the aggregator is case sensitive 🤔

dataset/dataset.go
```diff
@@ -48,6 +49,20 @@ type Client struct {
...
+	}
+}
+// closeResponseBody closes the response body and logs an error if unsuccessful
+
+func closeResponseBody(resp *http.Response) {
```
eldeal: do we really need the second function? as it's only used one, I'd suggest only having `func closeResponseBody(ctx, resp)` and just passing nil context in when it's not available? or `context.Background()`

dataset/dataset.go
dandwal: Haha, in which case this has been a mistake on all client updates.

PR #10, opened 2019-09-19
Enable filter client to accept user auth token

dandwal:
### What

* userAuthToken added to filter api client
* renamed testing vars to be inline with other client tests
* added nil body close check

### Who can review

Anyone except me

### DIFF
--- a/filter/filter.go
+++ b/filter/filter.go
@@ -86,7 +86,7 @@ func (c *Client) Healthcheck() (string, error) {
 }
 
 // GetOutput returns a filter output job for a given filter output id
-func (c *Client) GetOutput(ctx context.Context, serviceAuthToken, downloadServiceToken, filterOutputID string) (Model, error) {
+func (c *Client) GetOutput(ctx context.Context, userAuthToken, serviceAuthToken, downloadServiceToken, filterOutputID string) (Model, error) {
 	uri := fmt.Sprintf("%s/filter-outputs/%s", c.url, filterOutputID)
 	var m Model
 	clientlog.Do(ctx, "retrieving filter output", service, uri)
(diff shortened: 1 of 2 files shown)


filter/filter.go
```diff
@@ -96,6 +99,7 @@ func (c *Client) GetOutput(ctx context.Context, serviceAuthToken, downloadServic
 		return m, err
 	}
 
+	common.AddUserHeader(req, userAuthToken)
```
jondewijones: I think this will be setting the wrong header for this value. `AddUserHeader` adds a `User-Identity` header, whereas we'd want to set the `X-Florence-Token` header.

Example:
https://github.com/ONSdigital/dp-api-clients-go/blob/master/dataset/dataset.go#L505

filter/filter.go
```diff
@@ -568,3 +542,32 @@ func (c *Client) GetPreview(ctx context.Context, serviceAuthToken, downloadServi
...
+func (c *Client) doGetWithAuthHeadersAndWithDownloadToken(ctx context.Context, userAuthToken, serviceAuthToken, downloadServiceAuthToken, collectionID, uri string) (*http.Response, error) {
+	req, err := http.NewRequest(http.MethodGet, uri, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	headers.SetCollectionID(req, collectionID)
+	headers.SetUserAuthToken(req, userAuthToken)
+	headers.SetServiceAuthToken(req, serviceAuthToken)
+	common.AddDownloadServiceTokenHeader(req, downloadServiceAuthToken)
```
daiLlew: I think you can use `headers.SetDownloadServiceToken` now.

PR #11, opened 2019-09-19
Feature/add user token collection id params to methods

jondewijones:
### What

- Add user auth token and collection id params to methods. 
-Add func for adding tokens to headers

### Who can review

Anyone

### DIFF
--- a/dataset/dataset.go
+++ b/dataset/dataset.go
@@ -5,14 +5,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/ONSdigital/go-ns/log"
 	"io/ioutil"
 	"net/http"
 	"sort"
 	"strings"
 
+	"github.com/ONSdigital/go-ns/log"
+
 	"github.com/ONSdigital/dp-api-clients-go/clientlog"
-	"github.com/ONSdigital/dp-rchttp"
+	rchttp "github.com/ONSdigital/dp-rchttp"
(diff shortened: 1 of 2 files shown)


dataset/dataset.go
```diff
@@ -82,21 +83,12 @@ func (c *Client) Healthcheck() (string, error) {
 }
 
 // Get returns dataset level information for a given dataset id
-func (c *Client) Get(ctx context.Context, serviceToken, id  string) (m Model, err error) {
+func (c *Client) Get(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, id string) (m Model, err error) {
```
daiLlew: I think we could probably rename `id` to something more descriptive.

dataset/dataset.go
```diff
@@ -325,23 +281,12 @@ func (c *Client) GetEditions(ctx context.Context, serviceToken, id string) (m []
 }
 
 // GetVersions gets all versions for an edition from the dataset api
-func (c *Client) GetVersions(ctx context.Context, serviceToken, downloadServiceToken, id, edition string ) (m []Version, err error) {
+func (c *Client) GetVersions(ctx context.Context, userAuthToken, serviceAuthToken, downloadServiceAuthToken, collectionID, id, edition string) (m []Version, err error) {
```
daiLlew: again I think `id` could be renamed to something more descriptive.

PR #12, opened 2019-09-20
added new headers pkg as new source of truth for common request headers

daiLlew:
### What

Added new `headers` package to replace duplicated & inconsistent header utilities in `go-ns`.
Added functions for `Get/Set` `Collection-Id` header plus unit tests.

### DIFF
--- a/headers/headers.go
+++ b/headers/headers.go
@@ -38,8 +38,10 @@ func setRequestHeader(req *http.Request, headerName string, headerValue string) 
 		return errRequestNil
 	}
 
+	logD := log.Data{"header_name": headerName}
+
 	if len(headerValue) == 0 {
-		log.Event(context.Background(), "request header not set as value was empty", log.Data{"header_name": headerName})
+		log.Event(context.Background(), "request header not set as value was empty", logD)
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


headers/headers.go
```diff
@@ -0,0 +1,75 @@
+// headers package provides header name constants and get/set functions for commonly used http headers in the
+// dp-api-clients-go repo. Package replaces go-ns lib and should be treated as the single source of truth
```
dandwal: Likely worth us updating the readme in go-ns to mention this too

headers/headers.go
```diff
@@ -0,0 +1,75 @@
...
+
+	headerValue := req.Header.Get(headerName)
+	if len(headerValue) == 0 {
+		return "", ErrHeaderNotFound
+	}
+
+	return headerValue, nil
+}
+
+func setRequestHeader(req *http.Request, headerName string, headerValue string) error {
```
dandwal: Just a suggestion not an issue.
Alternativly could have
```suggestion
func setRequestHeader(req *http.Request, headerName, headerValue string) error {
```

daiLlew: Personally not a fan of that style

headers/headers.go
daiLlew: Its next on my list 😉

PR #13, opened 2019-09-20
added get/set functions for user auth token header plus unit tests

daiLlew:
Added functions to headers package to get/set the user auth token header plus unit tests.

### DIFF
--- a/headers/headers.go
+++ b/headers/headers.go
@@ -10,15 +10,46 @@ import (
 	"github.com/ONSdigital/log.go/log"
 )
 
-var (
-	// CollectionIDKey is the name used for a collection ID http request header
-	CollectionIDKey = "Collection-Id"
+const (
+	// CollectionIDHeader is the name used for a collection ID http request header
+	CollectionIDHeader = "Collection-Id"
 
+	// UserAuthTokenHeader is the user Florence auth token header name
+	UserAuthTokenHeader = "X-Florence-Token"
+)
+
+var (
 	// ErrHeaderNotFound returned if the requested header is not present in the provided request
 	ErrHeaderNotFound = errors.New("header not found")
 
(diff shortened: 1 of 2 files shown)


headers/headers.go
```diff
@@ -10,16 +10,47 @@ import (
 	"github.com/ONSdigital/log.go/log"
 )
 
-var (
-	// CollectionIDKey is the name used for a collection ID http request header
-	CollectionIDKey = "Collection-Id"
+const (
+	// CollectionIDHeader is the name used for a collection ID http request header
+	CollectionIDHeader = "Collection-Id"
```
daiLlew: Do we think that the header names should be private? That will help prevent people accidentally duplicating these functions.

headers/headers.go
```diff
@@ -10,16 +10,47 @@ import (
...
+
+	// UserAuthTokenHeader is the user Florence auth token header name
+	UserAuthTokenHeader = "X-Florence-Token"
```
daiLlew: Opinionated view: I think from now on this header should have a single name and all references to it should use that same name. This should help avoid confusion. I've gone with `UserAuthTokenHeader` but happy to argue over a different name.

headers/headers.go
dandwal: If there isn't a benefit for them being public then I would think private. I think keep accessors as limiting as they can be is a good practice to follow, nothing public that doesn't need to be etc.

So IMO Yes.

headers/headers.go
dandwal: Again just my personal view, I think `UserAuthToken` would suffice as these are all headers. I think as the package name is explicit and accurate for what it holds we don't need to append the names with header in this instance (so long as the content of this package doesn't ever get out of scope of itself).

I strongly agree about us setting single name and doing it now.

PR #16, opened 2019-09-30
Feature/header req

daiLlew:
Added `SetRequestID` & `GetRequestID ` and unit tests.

### DIFF
--- a/headers/headers_test.go
+++ b/headers/headers_test.go
@@ -394,6 +394,79 @@ func TestSetUserIdentity(t *testing.T) {
 	execSetHeaderTestCases(t, cases)
 }
 
+func TestSetRequestID(t *testing.T) {
+	cases := []setHeaderTestCase{
+		{
+			description: "SetRequestID should return error if request is nil",
+			getRequestFunc: func() *http.Request {
+				return nil
+			},
+			setHeaderFunc: func(r *http.Request) error {
+				return SetRequestID(r, "")
+			},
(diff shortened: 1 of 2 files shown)


headers/headers_test.go
```diff
@@ -394,6 +394,79 @@ func TestSetUserIdentity(t *testing.T) {
...
+		{
+			description: "SetRequestID should overwrite an existing header",
+			getRequestFunc: func() *http.Request {
+				return getRequestWithHeader(requestIDHeader, testHeader1)
+			},
+			setHeaderFunc: func(r *http.Request) error {
+				return SetRequestID(r, testHeader2)
+			},
+			getHeaderFunc: func(r *http.Request) string {
+				return r.Header.Get(requestIDHeader)
```
gedge: fyi - `Get` returns the _first_ value for that header (internally, it's an array), so you are leveraging that the original header is either:

1. replaced (which is the case using `Set`)
2. or that the array of values for a header is _appended to_ (which it is, if you were using `Add`), so the original one (returned) wouldn't change, but you're testing for that

so the test isn't wrong, but it does make assumptions that I want to ensure you're aware of

daiLlew: The library is a wrapper around the get/set header methods it doesn't change the functionality of `http.Request`. All it does is provide a convenient way to set/get one of our headers ensuring the correct header name value is used consistently.

If someone were to use `set` then later `add` I would expect the GetX functions here to behave the same as the underlying get in the http library.

PR #20, opened 2019-10-08
added helper func for checking if an error is ErrHeaderNotFound

daiLlew:
Added new func for checking if an error is equal to `ErrHeaderNotFound` to reduce the burden on the caller. 
A fairly common pattern in Go for "expected errors" i.e. `os.IsNotExist(err error)`. Added unit tests for new functionality.

### DIFF
--- a/headers/headers.go
+++ b/headers/headers.go
@@ -45,6 +45,11 @@ var (
 	ErrRequestNil = errors.New("error setting request header request was nil")
 )
 
+// IsNotFound return true if the err equals ErrHeaderNotFound return false otherwise
+func IsNotFound(err error) bool {
+	return err == ErrHeaderNotFound
+}
+
 // GetCollectionID returns the value of the "Collection-Id" request header if it exists, returns ErrHeaderNotFound if
 // the header is not found.
 func GetCollectionID(req *http.Request) (string, error) {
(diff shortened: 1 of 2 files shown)


headers/headers.go
```diff
@@ -45,6 +45,11 @@ var (
 	ErrRequestNil = errors.New("error setting request header request was nil")
 )
 
+// IsNotFound return true if the err equals ErrHeaderNotFound return false otherwise
+func IsNotFound(err error) bool {
+	return err == ErrHeaderNotFound
```
janderson2: Would it be considered better practice to use https://golang.org/pkg/errors/#Is.  That way if the error is wrapped it will still return true?

daiLlew: I've never seen that before - I am pretty sure we don't use that anywhere in any of our apps. Although now I question why not.

daiLlew: I tried to implement this but couldn't find it anywhere even after updating to the latest version of things. After some digging it appears that `errors.Is()` is newly added in **GO 1.13** and not in **1.12**. 
I think we should use this going forward but it requires us to update to 1.13 first. For now I will leave my PR unchanged will the view of updating it when we officially migrate to 1.13

PR #23, opened 2019-10-18
updated identity client to migrate from common -> headers

daiLlew:
### What

Updated identity client to migrate from old `common` headers to new `headers` package. And updated tests

### How to review

Code review, run locally, run tests.

### Who can review

Anyone else.

### DIFF
--- a/identity/authentication.go
+++ b/identity/authentication.go
@@ -100,25 +100,22 @@ func (api Client) CheckRequest(req *http.Request, florenceToken, serviceAuthToke
 		return ctx, resp.StatusCode, errors.WithMessage(errUnableToIdentifyRequest, "unexpected status code returned from AuthAPI"), nil
 	}
 
-	identityResp, err := unmarshalIdentityResponse(resp)
+	identityResponse, err := unmarshalIdentityResponse(resp)
 	if err != nil {
-		log.Event(ctx, "error unmarshaling AuthAPI identity response body", logData, log.Error(err))
 		return ctx, http.StatusInternalServerError, nil, err
 	}
 
-	var userIdentity string
-	if isUserReq {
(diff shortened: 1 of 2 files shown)


identity/authentication.go
```diff
@@ -107,7 +108,7 @@ func (api Client) CheckRequest(req *http.Request, florenceToken, serviceAuthToke
 	if isUserReq {
 		userIdentity = identityResp.Identifier
 	} else {
-		userIdentity = req.Header.Get(common.UserHeaderKey)
+		userIdentity, _  = headers.GetUserIdentity(req)
```
dandwal: Should we be throwing away the error without doing anything here?

identity/authentication.go
```diff
@@ -145,6 +145,32 @@ func splitToken(token string) (tokenObj tokenObject) {
 	return tokenObj
 }
 
+func createUserAuthRequest(url string, userAuthToken string) (*http.Request, error) {
```
dandwal: Only minor but think this function is missing an annotation, although it is a self-explaining function name, other functions in the file have done this, so think we should do this here too.

identity/authentication.go
```diff
@@ -145,6 +145,32 @@ func splitToken(token string) (tokenObj tokenObject) {
...
+	}
+
+	if err := headers.SetUserAuthToken(outboundAuthReq, userAuthToken); err != nil {
+		return nil, err
+	}
+
+	return outboundAuthReq, nil
+}
+
+func createServiceAuthRequest(url string, serviceAuthToken string) (*http.Request, error) {
```
dandwal: Likewise here

> Only minor but think this function is missing an annotation, although it is a self-explaining function name, other functions in the file have done this, so think we should do this here too.

PR #26, opened 2019-10-24
Fix query param encoding

jondewijones:
### What

Fix query param encoding.

### Who can review

Anyone but me.

### DIFF
--- a/zebedee/client.go
+++ b/zebedee/client.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"strconv"
 	"sync"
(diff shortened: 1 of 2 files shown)


zebedee/client.go
```diff
@@ -239,26 +241,27 @@ func (c *ZebedeeClient) GetPageTitle(ctx context.Context, userAccessToken, uri s
...
 		if !ok {
 			log.ErrorCtx(ctx, errors.New("error casting collection ID cookie to string"), nil)
 		}
-		url += "/" + collectionID
+		path += "/" + collectionID
 	}
 
-	url += "?" + query
+	path += "?" + url.PathEscape(query)
+	log.InfoCtx(ctx, "built uri", log.Data{"request_url": path})
```
janderson2: We might want to remove this debug log line.

PR #27, opened 2019-12-10
Add health checker function

nshumoogum:
### What

Add health checker function for calling API health endpoints and it abides the healthcheck.checker interface.

### How to review

Check code logic - particularly the creation of check object determined by the response from `/health` or `/healthcheck` endpoints, see [documentation](https://github.com/ONSdigital/dp/blob/master/standards/HEALTH_CHECK_SPECIFICATION.md#check-statuses)

### Who can review

Anyone

### DIFF
--- a/health/health.go
+++ b/health/health.go
@@ -27,8 +27,8 @@ type Client struct {
 	url    string
 }
 
-// New creates a new instance of Client with a given api url
-func New(name, url string) *Client {
+// NewClient creates a new instance of Client with a given api url
+func NewClient(name, url string) *Client {
 	c := &Client{
 		client: rchttp.NewClient(),
 		name:   name,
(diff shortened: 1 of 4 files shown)


health/health.go
```diff
@@ -0,0 +1,93 @@
...
+func New(name, url string) *Client {
+	c := &Client{
+		client: rchttp.NewClient(),
+		name:   name,
+		url:    url,
+	}
+
+	// Set the number of max retries to 1, allow caller to
+	// handle re-calling health checker
+	c.client.SetMaxRetries(1)
```
daiLlew: I wonder if accepting the max retries as a param to the function might be better? The logic here makes sense but it's assuming that this is always the case. By passing it as a parameter you are delegating that decision to the caller.

PR #31, opened 2020-01-17
Feature/update dataset list model

jondewijones:
### What

- Update dataset list model to include current and next when in publishing mode

### Who can review

Anyone


dataset/data.go
```diff
@@ -6,33 +6,40 @@ import (
...
-	State             string           `json:"state"`
-	Theme             string           `json:"theme"`
-	Title             string           `json:"title"`
-	UnitOfMeasure     string           `json:"unit_of_measure"`
-	URI               string           `json:"uri"`
+// DatasetDetails represents a response dataset model from the dataset api
+type DatasetDetails struct {
+	ID                string           `json:"id,omitempty"`
+	CollectionID      string           `json:"collection_id,omitempty"`
+	Contacts          []Contact        `json:"contacts,omitempty"`
```
nshumoogum: Any arrays which are not pointers (e.g. `*[]Contact`) will be returned as empty arrays in the json instead of not appearing at all. Thought I'd mention it as I don't know whether you mean to have empty arrays returned in the response?

jondewijones: 👍

dataset/data.go
```diff
@@ -6,33 +6,40 @@ import (
...
+// DatasetDetails represents a response dataset model from the dataset api
+type DatasetDetails struct {
+	ID                string            `json:"id,omitempty"`
+	CollectionID      string            `json:"collection_id,omitempty"`
+	Contacts          *[]Contact        `json:"contacts,omitempty"`
+	Description       string            `json:"description,omitempty"`
+	Keywords          *[]string         `json:"keywords,omitempty"`
+	License           string            `json:"license,omitempty"`
+	Links             Links             `json:"links,omitempty"`
+	Methodologies     *[]Methodology    `json:"methodologies,omitempty,omitempty"`
```
nshumoogum: Duplicate omitempty tag

PR #32, opened 2020-01-17
Update clients to include Checker function, replaces the old healthcheck func

nshumoogum:
### What

Add checker functions for current clients:

codelist
dataset
filter
hierarchy
importapi
search
zebedee

Update client packages to use latest version of the dp-rchttp library and switch out go-ns/log for log.go/log lib. Implement go modules for dependency management as well as using latest dp-healthcheck library in health package and add documentation.

### How to review

Check changes make sense and are logical
Check unit tests pass (and that it builds dependencies from go.mod)
Check all clients in library have Checker function and abides to healthcheck spec

### Who can review
Anyone


renderer/renderer.go
```diff
@@ -46,7 +46,7 @@ func New(url string) *Renderer {
 
 func closeResponseBody(ctx context.Context, resp *http.Response) {
 	if err := resp.Body.Close(); err != nil {
-		log.ErrorCtx(ctx, err, log.Data{"message": "error closing http response body"})
+		log.Event(ctx, "error closing http response body", log.Error(err))
 	}
 }
```
DavidSubiros: I see that renderer had a Healthcheck() function, but it doesn't have a Checker(). Do we need to implement a Checker function, and add a Check in the Renderer struct, similarly to what you did for other services?

health/health.go
```diff
@@ -3,90 +3,134 @@ package health
...
+	c.Check.StatusCode = code
+	c.Check.LastChecked = &currentTime
+
+	switch code {
+	case 200:
+		c.Check.Message = StatusMessage[health.StatusOK]
+		c.Check.Status = health.StatusOK
+		c.Check.LastSuccess = &currentTime
+	case 429:
+		c.Check.Message = StatusMessage[health.StatusWarning]
```
DavidSubiros: Do we want the default messages for each code, or shall we use error.Error() ?

health/health.go
```diff
@@ -3,90 +3,134 @@ package health
...
+	"github.com/ONSdigital/dp-api-clients-go/clientlog"
 	health "github.com/ONSdigital/dp-healthcheck/healthcheck"
 	rchttp "github.com/ONSdigital/dp-rchttp"
 	"github.com/ONSdigital/log.go/log"
 )
 
-// ErrInvalidAPIResponse is returned when an api does not respond
+var (
+	// StatusMessage contains a map of messages to service response statuses
+	StatusMessage = map[string]string{
```
DavidSubiros: Do we want to use default messages? That would shadow the error that caused a warning or a critical, but on the other hand, this would make the error concise.

renderer/renderer.go
nshumoogum: Good spot

health/health.go
nshumoogum: Simply put yes, I've gone back and forth on this but when I went through the docs with James he referred to the [documentation](https://github.com/ONSdigital/dp/blob/master/standards/HEALTH_CHECK_SPECIFICATION.md#check-statuses) specifically for api clients that only act on status codes

health/health.go
nshumoogum: It is not for an applications health to indicate the reason to why another app is failing; just to indicate there is an issue with an upstream application.

The only case where I think you are right to return the error in the message is if there was an issue with the client; so I have updated the health.Checker function to return this error if it occurs.

health/health.go
nshumoogum: See comment below

PR #44, opened 2020-02-28
Feature/check token identity

DavidSubiros:
### What

- identity package:
  - Created `CheckTokenIdentity` in identity package. This method validates a token (service and user tokens supported) against `<zebedeeURL>/identity`.
  - Refactor existing method `CheckRequest` to use `CheckTokenIdentity`, which does the same plus some extra functionality after validating the token.

- dataset package:
  -  Constructor with max retries `NewAPIClientWithMaxRetries `, so callers can define max retries for the HTTP Client.
  - Added ID to Dimension model, as expected by dp-dimension-extractor to be returned in the array of Dimensions returned by `GetInstance`
  - Implemented `PostInstanceDimensions` to perform `POST /instances/<id>/dimensions`
  - Implemented `PutInstanceData` to perform `PUT /instances/<id>`

### How to review

- Unit tests should pass and cover enough cases of the new code
- Make sure the code changes make sense (keeping in mind that they are migrated from dp-dimension-extractor, but might be useful for other services)

### Who can review

Anyone.

### DIFF
--- a/dataset/dataset.go
+++ b/dataset/dataset.go
@@ -359,6 +359,29 @@ func (c *Client) GetInstance(ctx context.Context, userAuthToken, serviceAuthToke
 	return
 }
 
+// PostInstanceDimensions performs a 'POST /instances/<id>/dimensions' with the provided OptionPost
+func (c *Client) PostInstanceDimensions(ctx context.Context, serviceAuthToken, instanceID string, optionToPost OptionPost) error {
+	payload, err := json.Marshal(optionToPost)
+	if err != nil {
+		return err
+	}
+
+	uri := fmt.Sprintf("%s/instances/%s/dimensions", c.url, instanceID)
+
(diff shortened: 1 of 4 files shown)


dataset/dataset.go
```diff
@@ -68,6 +68,18 @@ func NewAPIClient(datasetAPIURL string) *Client {
 	}
 }
 
+// NewAPIClientWithMaxRetries creates a new instance of Client with a given dataset api url and the relevant tokens,
+// setting a number of max retires for the HTTP client
+func NewAPIClientWithMaxRetries(datasetAPIURL string, maxRetries int) *Client {
```
nshumoogum: I assume dimension extractor needed a different retry count to the default?

Also remember `/health` and `/healthcheck` endpoints are whitelisted to only ever retry once

PR #46, opened 2020-03-04
Add observation importer dataset client methods

nshumoogum:
### What

Add additional methods that are needed from observation importer dataset client. This included returning bytes instead of struct defined by model in json marshaller as the bytes needed to be marshalled using a csv marshaller.

### How to review

* Check changes make sense
* Unit tests pass

### Who can review

Anyone


dataset/dataset.go
```diff
@@ -24,23 +24,23 @@ const service = "dataset-api"
 // ErrInvalidDatasetAPIResponse is returned when the dataset api does not respond
 // with a valid status
 type ErrInvalidDatasetAPIResponse struct {
-	actualCode int
+	ActualCode int
```
DavidSubiros: Do we need to export this field? (there is a getter function `Code()` which returns it)

dataset/dataset.go
```diff
@@ -350,12 +361,36 @@ func (c *Client) GetInstance(ctx context.Context, userAuthToken, serviceAuthToke
...
+	b, err = ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return
+	}
+
+	return
+}
+
+// GetInstanceDimensionsByBytes returns a list of dimensions for an instance as bytes from the dataset api
+func (c *Client) GetInstanceDimensionsByBytes(ctx context.Context, userAuthToken, serviceAuthToken, instanceID string) (b []byte, err error) {
```
DavidSubiros: In filterAPI client, I did something similar, but with the naming `...Bytes` instead of `...ByBytes`. Maybe it would be a good idea to have the same convention? I can modify FilterAPI if we want to use `ByBytes` or we can rename this function to `GetInstanceDimensionsBytes`

nshumoogum: I'll change it to meet your convention

dataset/dataset.go
nshumoogum: I need to set this value to different status codes as observation importer extracts the code to handle the error message it returns within application, so this can be exported or I create a SetCode method

PR #51, opened 2020-05-21
Add image-api client

jessjenkins:
### What

Added image-api client

### How to review

Ensure that endpoints defined in [dp-image-api/swagger.yaml](https://github.com/ONSdigital/dp-image-api/blob/develop/swagger.yaml) have been implemented and tested.

### Who can review

Anyone but me

### DIFF
--- a/image/data.go
+++ b/image/data.go
@@ -1,0 +1,59 @@
+package image
+
+type Images struct {
+	Count      int     `json:"count"`
+	Items      []Image `json:"items"`
+	Limit      int     `json:"limit"`
+	Offset     int     `json:"offset"`
+	TotalCount int     `json:"total_count"`
+}
+
+type NewImage struct {
+	CollectionId string  `json:"collection_id,omitempty"`
+	Filename string  `json:"filename,omitempty"`
+	License License   `json:"license,omitempty"`
+	Type string  `json:"type,omitempty"`
+}
+
+type License struct {
+	Title string  `json:"title,omitempty"`
+	Href string  `json:"href,omitempty"`
+}
+
+type Image struct {
+	Id string  `json:"id,omitempty"`
+	CollectionId	 string  `json:"collection_id,omitempty"`
+	State string  `json:"state,omitempty"`
+	//enum:
+	//- created
+	//- uploaded
+	//- publishing
+	//- published
+	//- deleted
+	Filename string  `json:"filename,omitempty"`
+	License License   `json:"license,omitempty"`
+	Upload ImageUpload `json:"upload,omitempty"`
+	Type string  `json:"type,omitempty"`
+	Downloads ImageDownloads  `json:"downloads,omitempty"`
+}
+
+type ImageUpload struct {
+	Path string  `json:"path,omitempty"`
+}
+
+type ImageDownloads struct {
+	Png ImagedDownloadsVariant  `json:"png,omitempty"`
+}
+
+type ImagedDownloadsVariant struct {
+	Size1920x1080 ImageDownload  `json:"1920x1080,omitempty"`
+	Size1280x720 ImageDownload  `json:" 1280x720,omitempty"`
+	ThumbNail ImageDownload  `json:"thumbnail,omitempty"`
+}
+
+type ImageDownload struct {
+	Size int `json:"size,omitempty"`
+	Href string `json:"href,omitempty"`
+	Public string `json:"public,omitempty"`
+	Private string `json:"private,omitempty"`
+}
--- a/image/data.go
+++ b/image/data.go
@@ -10,50 +10,40 @@ type Images struct {
 
 type NewImage struct {
 	CollectionId string  `json:"collection_id,omitempty"`
-	Filename string  `json:"filename,omitempty"`
-	License License   `json:"license,omitempty"`
-	Type string  `json:"type,omitempty"`
+	Filename     string  `json:"filename,omitempty"`
+	License      License `json:"license,omitempty"`
+	Type         string  `json:"type,omitempty"`
 }
 
 type License struct {
-	Title string  `json:"title,omitempty"`
-	Href string  `json:"href,omitempty"`
+	Title string `json:"title,omitempty"`
+	Href  string `json:"href,omitempty"`
 }
 
 type Image struct {
-	Id string  `json:"id,omitempty"`
-	CollectionId	 string  `json:"collection_id,omitempty"`
-	State string  `json:"state,omitempty"`
+	Id           string `json:"id,omitempty"`
+	CollectionId string `json:"collection_id,omitempty"`
+	State        string `json:"state,omitempty"`
 	//enum:
 	//- created
 	//- uploaded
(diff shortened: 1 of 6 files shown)


image/data.go
```diff
@@ -0,0 +1,59 @@
...
+type Image struct {
+	Id string  `json:"id,omitempty"`
+	CollectionId	 string  `json:"collection_id,omitempty"`
+	State string  `json:"state,omitempty"`
+	//enum:
+	//- created
+	//- uploaded
+	//- publishing
+	//- published
+	//- deleted
```
jessjenkins, line 32: I am not sure if anything should be done with enums in clients. I have temporarily left the comments in place for discussion in this PR.

image/image_test.go
```diff
@@ -0,0 +1,479 @@
...
+			m, err := cli.GetImages(ctx)
+
+			Convey("a positive response is returned", func() {
+				So(err, ShouldBeNil)
+				So(m.Count, ShouldEqual, 1)
+				So(m.Items, ShouldNotBeEmpty)
+				So(m.Items, ShouldHaveLength, 1)
+				mItem := m.Items[0]
+				So(mItem.Id, ShouldResemble, "042e216a-7822-4fa0-a3d6-e3f5248ffc35")
+				So(mItem.Downloads, ShouldNotBeEmpty)
```
jessjenkins, line 291: NB. I've added a few checks on the object, but I'm not convinced that the unit test should be checking every field as that's really testing `json.Unmarshal` then rather than this library. I can add more if necessary but not sure there's much value.

image/data.go
```diff
@@ -0,0 +1,59 @@
...
+	Path string  `json:"path,omitempty"`
+}
+
+type ImageDownloads struct {
+	Png ImagedDownloadsVariant  `json:"png,omitempty"`
+}
+
+type ImagedDownloadsVariant struct {
+	Size1920x1080 ImageDownload  `json:"1920x1080,omitempty"`
+	Size1280x720 ImageDownload  `json:" 1280x720,omitempty"`
```
jessjenkins, line 50: NB. This conforms to the swagger spec, but I'm not convinced this is the right approach. We need to discuss the keys for image variants in more detail.

image/data.go
DavidSubiros: In dataset client, I used iota constants for the state enumeration. Maybe we can do something similar?:
```
// Possible values for a State of the resource. It can only be one of the following:
// TODO these states should be enforced in all the 'POST' and 'PUT' operations that can modify states of resources
const (
	StateCreated State = iota
	StateUploaded
	StatePublishing
	StatePublished
	StateDeleted
)

var stateValues = []string{"created", "uploaded", "publishing", "published", "deleted"}

// String returns the string representation of a state
func (s State) String() string {
	return stateValues[s]
}
```

image/data.go
```diff
@@ -0,0 +1,59 @@
...
+	Upload ImageUpload `json:"upload,omitempty"`
+	Type string  `json:"type,omitempty"`
+	Downloads ImageDownloads  `json:"downloads,omitempty"`
+}
+
+type ImageUpload struct {
+	Path string  `json:"path,omitempty"`
+}
+
+type ImageDownloads struct {
```
DavidSubiros, line 44: If we use generic maps, we could delete this struct

image/data.go
```diff
@@ -0,0 +1,59 @@
...
+	//- created
+	//- uploaded
+	//- publishing
+	//- published
+	//- deleted
+	Filename string  `json:"filename,omitempty"`
+	License License   `json:"license,omitempty"`
+	Upload ImageUpload `json:"upload,omitempty"`
+	Type string  `json:"type,omitempty"`
+	Downloads ImageDownloads  `json:"downloads,omitempty"`
```
DavidSubiros, line 37: Suggestion described below, about using generic maps?
```
Downloads map[string]map[string]ImageDownload `json:"downloads,omitempty"`
```

image/data.go
```diff
@@ -0,0 +1,59 @@
...
+
+type ImageUpload struct {
+	Path string  `json:"path,omitempty"`
+}
+
+type ImageDownloads struct {
+	Png ImagedDownloadsVariant  `json:"png,omitempty"`
+}
+
+type ImagedDownloadsVariant struct {
```
DavidSubiros, line 48: if we used generic maps we could delete this struct

image/data.go
DavidSubiros: Maybe we can use a generic nested map in ImageDownloads, to allow for generic extensions and resolutions (or some other description like `thumbnail`)?
Unless we define a rigid set of possible keys, but I think we don't have a clear use case yet.

image/image.go
```diff
@@ -0,0 +1,255 @@
...
+	if err = json.Unmarshal(b, &m); err != nil {
+		return
+	}
+
+	return
+}
+
+// PostImage performs a 'POST /images' with the provided NewImage
+func (c *Client) PostImage(ctx context.Context, data NewImage) (m Image, err error) {
+	payload, err := json.Marshal(data)
```
DavidSubiros, line 129: If we decide to use the iota enum, here we could verify that the provided Image state is a valid state

image/image.go
```diff
@@ -0,0 +1,255 @@
...
+	if err = json.Unmarshal(b, &m); err != nil {
+		return
+	}
+
+	return
+}
+
+// PutImage updates the specified image
+func (c *Client) PutImage(ctx context.Context, imageID string, data Image) (m Image, err error) {
+	payload, err := json.Marshal(data)
```
DavidSubiros, line 202: If we decide to use the iota enum, here we could verify that the provided Image state is a valid state

image/image_test.go
DavidSubiros: I agree that we would be testing the library, maybe it's not necessary to add more tests which would be actually testing the serialisation.

image/image_test.go
```diff
@@ -0,0 +1,479 @@
...
+			})
+
+			Convey("and dphttpclient.Do is called 1 time", func() {
+				checkResponseBase(mockdphttpCli, http.MethodGet, "/images")
+			})
+		})
+	})
+}
+
+func TestClient_PostImage(t *testing.T) {
```
DavidSubiros, line 304: If we decide to use the iota enum, here we could add a test with an invalid state

image/image_test.go
```diff
@@ -0,0 +1,479 @@
...
+			})
+
+			Convey("and dphttpclient.Do is called 1 time", func() {
+				checkResponseBase(mockdphttpCli, http.MethodGet, "/images/123")
+			})
+		})
+	})
+}
+
+func TestClient_PutImage(t *testing.T) {
```
DavidSubiros, line 414: If we decide to use the iota enum, here we could add a test with an invalid state

image/data.go
DavidSubiros: I did state validation in the server, I think I should not block this PR because of this reason - I will approve it.

PR #52, opened 2020-06-10
Added auth toknes to image api

DavidSubiros:
### What

For Image API client:
- Added headers for service auth, user auth and collectionID
- Modified unit tests to verify headers for all calls 

### How to review

- Make sure unit tests pass

### Who can review

Anyone

### DIFF
--- a/image/image.go
+++ b/image/image.go
@@ -244,9 +244,9 @@ func addCollectionIDHeader(r *http.Request, collectionID string) {
 	}
 }
 
-// doGetWithAuthHeaders executes clienter.Do setting the user and service authentication token as a request header. Returns the http.Response and any error.
-// It is the callers responsibility to ensure response.Body is closed on completion.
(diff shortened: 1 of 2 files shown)


image/image.go
```diff
@@ -253,3 +237,52 @@ func NewImageAPIResponse(resp *http.Response, uri string) (e *ErrInvalidImageAPI
...
 	return
 }
+
+func addCollectionIDHeader(r *http.Request, collectionID string) {
+	if len(collectionID) > 0 {
+		r.Header.Add(common.CollectionIDHeaderKey, collectionID)
+	}
+}
+
+// doGetWithAuthHeaders executes clienter.Do setting the user and service authentication token as a request header. Returns the http.Response and any error.
```
jessjenkins, line 247: This is the only method here that has a comment. Also the comment mentions the auth tokens but not the collection header. I suggest either completing the other methods/headers or else deciding to remove the comment so there's consistency.

PR #62, opened 2020-07-17
Investigate re-usage of Clienters among different API clients.

DavidSubiros:
### What

We want to have the capability to re-use a dp-net/http Clienter among different API clients. In this PR, there is the suggested implementation in Image API client, that all clients should follow.

### How to review

- Make sure that the new constructor function in ImageAPI and Health clients allows a Clienter to be reused
- Check if the changes in the README are clear
- Make sure unit tests pass

### Who can review

Anyone

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,20 +19,25 @@ Common client code - in go - for ONS APIs:
 
 #### Usage
 
-(WIP) Each client defines two constructor functions: one that creates a new dp-net/http Clienter, and the other that allows you to provide it externally, so that you can reuse it among different clients.
+(WIP) Each client defines two constructor functions: one that creates a new healthcheck client (with a new dp-net/http Clienter), and the other that allows you to provide it externally, so that you can reuse it among different clients.
 
-For example, you may initialise an image API client with a new Clienter:
+For example, you may create a new image API client like so:
 ```
     import  "github.com/ONSdigital/dp-api-clients-go/image"
+
     ...
     imageClient := image.NewAPIClient(<url>)
     ...
 ```
 
-Or you may initialise it providing a Clienter:
+Or you may create it providing a Healthcheck client:
 ```
+    import  "github.com/ONSdigital/dp-api-clients-go/image"
+    import  "github.com/ONSdigital/dp-api-clients-go/health"
+
     ...
-    imageClient := image.NewWithClienter(<url>, <clienter>)
+    hcClient := health.NewClient(<genericName>, <url>)
+    imageClient := image.NewWithHealthClient(hcClient)
     ...
 ```
 
--- a/README.md
+++ b/README.md
@@ -11,9 +11,30 @@ Common client code - in go - for ONS APIs:
 * healthcheck -> health
 * hierarchy
 * identity
+* image
 * importapi
 * renderer
 * search
+
+
+#### Usage
+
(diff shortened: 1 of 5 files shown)


health/health.go
```diff
@@ -37,8 +37,13 @@ type Client struct {
 
 // NewClient creates a new instance of Client with a given app url
 func NewClient(name, url string) *Client {
```
CarlHembrough, line 39: I wonder if it would be better to remove the original constructors to force the consideration of clienter reuse, at the expense of the convenience? 🤔

image/image.go
```diff
@@ -58,44 +57,39 @@ func closeResponseBody(ctx context.Context, resp *http.Response) {
...
 
-// NewAPIClient creates a new instance of Client with a given image api url and the relevant tokens
+// NewAPIClient creates a new instance of ImageAPI Client with a given image api url
 func NewAPIClient(imageAPIURL string) *Client {
-	hcClient := healthcheck.NewClient(service, imageAPIURL)
-
 	return &Client{
-		cli: hcClient.Client,
-		url: imageAPIURL,
+		healthcheck.NewClient(service, imageAPIURL),
```
CarlHembrough, line 63: I think this is why I thought it odd to have the option to pass in the healthcheck. Would it be better to always initialise the healthcheck client internally to the client? Then instead of creating the healthcheck client to pass in, you would create and pass in the clienter.

README.md
```diff
@@ -11,10 +11,36 @@ Common client code - in go - for ONS APIs:
...
 * identity
+* image
 * importapi
 * renderer
 * search
 
+
+#### Usage
+
+(WIP) Each client defines two constructor functions: one that creates a new healthcheck client (with a new dp-net/http Clienter), and the other that allows you to provide it externally, so that you can reuse it among different clients.
```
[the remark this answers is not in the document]
DavidSubiros: Yes, maybe it's confusing - I put `WIP` here because most of the clients don't follow this, yet.

README.md
```diff
@@ -11,10 +11,36 @@ Common client code - in go - for ONS APIs:
...
+```
+
+Or you may create it providing a Healthcheck client:
+```
+    import  "github.com/ONSdigital/dp-api-clients-go/image"
+    import  "github.com/ONSdigital/dp-api-clients-go/health"
+
+    ...
+    hcClient := health.NewClient(<genericName>, <url>)
+    imageClient := image.NewWithHealthClient(hcClient)
```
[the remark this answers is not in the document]
DavidSubiros: We had a discussion with Eleonor about this: https://trello.com/c/R9Yl8Axr/4804-investigate-whether-we-can-consolidate-api-config-values-without-merging-clients-2d

 Originally, I created something like this:
 ```NewClientWithClienter(name, url string, clienter dphttp.Clienter) *Client```
You're right about maybe not being very clear. After the discussion about the URL (which would need to be provided in both cases), I did the change to pass the Health Client to the constructor instead of the clienter (so the API would be a kind of 'extension' of healthcheck client). My initial intention was to pass the clienter.

health/health.go
DavidSubiros: yes, that would be the way to make it clear that we want to share clienters among different clients.

PR #74, opened 2020-09-23
added public constructor for ErrInvalidHierarchyAPIResponse

daiLlew:
### What

Added public constructor for ErrInvalidHierarchyAPIResponse. This enables code using this library to create an instance of this error type. This is required when unit test code that uses the additional fields provided by this err type.

### DIFF
--- a/hierarchy/hierarchy.go
+++ b/hierarchy/hierarchy.go
@@ -125,7 +125,6 @@ func (c *Client) getHierarchy(ctx context.Context, path string) (Model, error) {
 	defer closeResponseBody(ctx, resp)
 
 	if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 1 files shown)


hierarchy/hierarchy.go
```diff
@@ -115,7 +125,8 @@ func (c *Client) getHierarchy(ctx context.Context, path string) (Model, error) {
 	defer closeResponseBody(ctx, resp)
 
 	if resp.StatusCode != http.StatusOK {
-		return m, &ErrInvalidHierarchyAPIResponse{http.StatusOK, resp.StatusCode, path}
+		//return m, &ErrInvalidHierarchyAPIResponse{http.StatusOK, resp.StatusCode, path}
```
nshumoogum, line 128: Commented out line probably needs to go

nshumoogum (changes requested):

CarlHembrough (approved):

PR #75, opened 2020-10-22
Add get resource body method

jondewijones:
### What

Add get resource body method

### Who can review

Anyone

### DIFF
--- a/zebedee/client.go
+++ b/zebedee/client.go
@@ -29,15 +29,15 @@ type Client struct {
 // ErrInvalidZebedeeResponse is returned when zebedee does not respond
 // with a valid status
 type ErrInvalidZebedeeResponse struct {
-	actualCode int
-	uri        string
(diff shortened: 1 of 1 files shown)


zebedee/client.go
```diff
@@ -282,6 +282,14 @@ func (c *Client) GetTimeseriesMainFigure(ctx context.Context, userAccessToken, c
 	return ts, nil
 }
 
+func (c *Client) GetResourceBody(ctx context.Context, userAccessToken, collectionID, lang, uri string) ([]byte, error) {
```
jessjenkins, line 285: Exported funcs should have godoc
```suggestion
// GetResourceBody returns …
func (c *Client) GetResourceBody(ctx context.Context, userAccessToken, collectionID, lang, uri string) ([]byte, error) {
```

janderson2 (approved):

PR #77, opened 2020-11-20
Feature/filter client patch

DavidSubiros:
### What

- Implemented PATCH client for new `AddDimensionOptions`, `RemoveDimensionOptions` and `PatchDeminsionOptions` functions
  - use dp-net request Patch struct.
- PATCH operations are done in batches of size `batchSize` (provided as a parameter to the function)
- PATCH operations are done in a single batch if the sum of items to add and remove is smaller than `batchSize`
- Added unit tests for the new functionality.
- Renamed existing `AddDimensionOptions` to `SetDimensionOptions`, because it overwites all items.

### How to review

- Make sure unit test are successful
- (Not needed for PR review, just information):
  - We tried to replace a mock with a real HTTP client, replace values tokens, etc., and run it against a filter-api running locally, and the behaviour was as expected.
  - We tried to use this branch in frontend-filter-dataset-controller and we observe the patch correctly sent to filter API.

### Who can review

Anyone


filter/filter_test.go
```diff
@@ -689,6 +689,65 @@ func TestClient_AddDimension(t *testing.T) {
...
+		Convey("when AddDimensionValues is called", func() {
+			err := filterClient.AddDimensionValues(ctx, testUserAuthToken, testServiceToken, testCollectionID, filterID, name, options)
+
+			Convey("then the expected error is returned", func() {
+				So(err.Error(), ShouldResemble, mockErr.Error())
+			})
+		})
+	})
+
+	Convey("given dphttpclient.do returns a non 200 response status", t, func() {
```
redhug1, line 826: What is '.do' after 'dphttpclient' ?

DavidSubiros: It performs an HTTP call

redhug1 (approved):
LGTM

PR #79, opened 2020-12-08
dataset GetOptions with offset and limit

DavidSubiros:
### What

- Modified dataset API client to support pagination for getOptions:
  - add limit and offset query params
  - modified returned model to include limit, offset, count and total count

### How to review

- Make sure unit tests pass

### Who can review

Anyone

### DIFF
--- a/dataset/dataset.go
+++ b/dataset/dataset.go
@@ -675,8 +675,8 @@ func (c *Client) GetVersionDimensions(ctx context.Context, userAuthToken, servic
 }
 
 // GetOptions will return the options for a dimension
-func (c *Client) GetOptions(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, id, edition, version, dimension string) (m Options, err error) {
-	uri := fmt.Sprintf("%s/datasets/%s/editions/%s/versions/%s/dimensions/%s/options", c.hcCli.URL, id, edition, version, dimension)
+func (c *Client) GetOptions(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, id, edition, version, dimension string, offset, limit int) (m Options, err error) {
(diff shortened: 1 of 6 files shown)


dataset/dataset.go
```diff
@@ -675,8 +675,8 @@ func (c *Client) GetVersionDimensions(ctx context.Context, userAuthToken, servic
 }
 
 // GetOptions will return the options for a dimension
-func (c *Client) GetOptions(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, id, edition, version, dimension string) (m Options, err error) {
-	uri := fmt.Sprintf("%s/datasets/%s/editions/%s/versions/%s/dimensions/%s/options", c.hcCli.URL, id, edition, version, dimension)
+func (c *Client) GetOptions(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, id, edition, version, dimension string, offset, limit int) (m Options, err error) {
```
daiLlew, line 678: Should we validate the new values to ensure nothing crazy is passed through? What happens if negative values are provided - that kind of stuff.

DavidSubiros: Good point, currently the dataset API replaces negative values with 0 (default) - do you think it should be a 400 Bad request? In that case, we could do the validation that you mention, and return an error here without doing the call

DavidSubiros: In terms of other crazy values (for example huge values), I think they shouldn't be a problem. If you request offset 999,999, any limit, and there are only 100 items, an empty list of items will be returned; or if you request an offset of 0 and a limit of 999,999 then only the 100 items will be returned.

DavidSubiros: Updated wit validation (negative values not allowed)

CarlHembrough (approved):

PR #80, opened 2020-12-16
Add put methods for datasets, versions and instances in dataset and Zebedee clients

jondewijones:
### What

Add put methods for datasets, versions and instances in dataset and Zebedee clients

### Who can review

Anyone


dataset/dataset.go
```diff
@@ -216,6 +216,29 @@ func (c *Client) GetDatasets(ctx context.Context, userAuthToken, serviceAuthToke
...
+	}
+
+	resp, err := c.doPutWithAuthHeaders(ctx, userAuthToken, serviceAuthToken, collectionID, uri, payload)
+	if err != nil {
+		return errors.Wrap(err, "http client returned error while attempting to make request")
+	}
+	defer closeResponseBody(ctx, resp)
+
+	if resp.StatusCode != http.StatusOK {
+		return errors.Errorf("incorrect http status, expected: 200, actual: %d, uri: %s", resp.StatusCode, uri)
```
DavidSubiros, line 237: other methods in dataset api client return a DatasetAPIResonse, I think it would be worth returning it here too, for consistency reasons:
```
 return NewDatasetAPIResponse(resp, uri)
```

dataset/dataset.go
```diff
@@ -446,6 +469,29 @@ func (c *Client) GetInstances(ctx context.Context, userAuthToken, serviceAuthTok
...
```
DavidSubiros, line 490: same comment as above, suggested :
```
return NewDatasetAPIResponse(resp, uri)
```

DavidSubiros (commented):
A couple of small comments on the errors being returned. Also, would it be possible to create some unit tests? (or maybe in another PR?)

dataset/dataset.go
jondewijones: 👍

DavidSubiros (approved):

PR #82, opened 2020-12-24
filter GetDimensions response is now mapped to the paginated struct

DavidSubiros:
### What

- Introduced Dimensions struct in filter client, to match the paginated response from filter API
- Used this struct as the return type for filter.GetDimensions
  - changed unit tests accordingly
- Use QueryParams struct to provide offset and limit to `GetDimensionOptions` and `GetDimensions`

### How to review

- Make sure unit tests pass

### Who can review

Anyone

### DIFF
--- a/filter/filter.go
+++ b/filter/filter.go
@@ -55,6 +55,20 @@ type Client struct {
 	hcCli *healthcheck.Client
 }
 
+// QueryParams represents the possible query parameters that a caller can provide
+type QueryParams struct {
+	Offset int
+	Limit  int
+}
+
+// Validate validates tht no negative values are provided for limit or offset
+func (q QueryParams) Validate() error {
+	if q.Offset < 0 || q.Limit < 0 {
+		return errors.New("negative offsets or limits are not allowed")
+	}
+	return nil
+}
+
 // New creates a new instance of Client with a given filter api url
 func New(filterAPIURL string) *Client {
 	return &Client{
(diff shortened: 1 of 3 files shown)


filter/filter.go
```diff
@@ -191,8 +205,8 @@ func (c *Client) GetDimensionBytes(ctx context.Context, userAuthToken, serviceAu
 }
 
 // GetDimensions will return the dimensions associated with the provided filter id as an array of Dimension structs
-func (c *Client) GetDimensions(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, filterID string) (dims []Dimension, err error) {
-	b, err := c.GetDimensionsBytes(ctx, userAuthToken, serviceAuthToken, collectionID, filterID)
+func (c *Client) GetDimensions(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, filterID string, q QueryParams) (dims Dimensions, err error) {
```
jessjenkins, line 208: Strictly speaking this should trigger a V2 release due to a breaking change. However being small, imposing a V2 change on all the other api clients would be disproportionately disruptive.

jessjenkins (approved):

filter/filter.go
DavidSubiros: Yes, I agree - the only affected service is `dp-frontend-filter-dataset-conroller`.

PR #85, opened 2021-01-07
Zebedee client can now be more configurable

dandwal:
### What

Timeout and maximum retries can now be set in the new constructor

#### Why is this needed?

Currently, there is an insane number of retries between the dp-frontend-router and Zebedee. From client level, web server level and platform level. As such we are able to remove the client level retries. Requests that are failing are staying open too long and this is hindering the dp-frontend-routers performance to the point it causes fall overs. This PR should help to address these issues in conjunction with other improvements.

### How to review

In dp-frontend-router or another service that uses this client run:
go mod edit -replace github.com/ONSdigital/dp-api-clients-go=/{path to your local dp-api-clients-go}

Replacing the {path to your local dp-api-clients-go} with your local path to your copy of the dp-api-clients-go directory.
Then checkout this PR branch.

Try a couple of pages of the website locally and ensure everything still works as expected.

To see how this might be used now in the future checkout this PR https://github.com/ONSdigital/dp-frontend-router/pull/247

### Who can review

Anyone except me

### DIFF
--- a/zebedee/client.go
+++ b/zebedee/client.go
@@ -68,12 +68,8 @@ func New(zebedeeURL string) *Client {
 }
 
 // NewWithSetTimeoutAndMaxRetry creates a new Zebedee Client, with a configurable timeout and maximum number of retries
-func NewWithSetTimeoutAndMaxRetry(zebedeeURL string, timeout, retry int) *Client {
-	hcClienter := dphttp.NewClient()
-
-	hcClient := healthcheck.NewClientWithClienter(service, zebedeeURL, hcClienter)
-	hcClient.Client.SetTimeout(time.Duration(timeout) * time.Second)
-	hcClient.Client.SetMaxRetries(retry)
+func NewClientWithClienter(zebedeeURL string, clienter dphttp.Clienter) *Client {
+	hcClient := healthcheck.NewClientWithClienter(service, zebedeeURL, clienter)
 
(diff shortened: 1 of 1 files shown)


zebedee/client.go
```diff
@@ -66,6 +67,19 @@ func New(zebedeeURL string) *Client {
 	}
 }
 
+// NewWithSetTimeoutAndMaxRetry creates a new Zebedee Client, with a configurable timeout and maximum number of retries
+func NewWithSetTimeoutAndMaxRetry(zebedeeURL string, timeout, retry int) *Client {
```
daiLlew, line 71: This is perfectly fine and will solve the problem we were discussing. However, I would personally do something like this instead:

```
func NewWithSetTimeoutAndMaxRetry(zebedeeURL string, clienter dphttp.Clienter) *Client {
        hcClient := healthcheck.NewClientWithClienter(service, zebedeeURL, clienter)

	return &Client{
		hcClient,
	}
}
```
Accepting the dphttp clienter as a param removes the internal set up of the client from the scope of this function - you give it the client you want it to use pre configured (retries and timeout values are set by the code passing it into this func. It should decouple it from the config and allow you to test it easier.

That would be my approach, I wont enforce that though  I'll leave it up to you to decided.

rav-pradhan (approved):

PR #88, opened 2021-01-22
As part of the query parameter validation, the provided IDs are escaped.

DavidSubiros:
### What

We identified a bug in develop, where some dimension options failed to be added to filters. After some investigation, we found that the issue was with URL encoding of IDs passed as query parameters when we were validating them.

This PR fixes this bug by escaping all the IDs as part of the query param validation.

### How to review

- Make sure unit tests pass and cover the new functionality

### Who can review

Anyone

### DIFF
--- a/dataset/dataset_test.go
+++ b/dataset/dataset_test.go
@@ -799,7 +799,7 @@ func TestClient_GetOptions(t *testing.T) {
 		datasetClient := newDatasetClient(httpClient)
 
 		Convey("when GetOptions is called with valid values for limit and offset", func() {
-			q := QueryParams{offset, limit, []string{}}
+			q := QueryParams{Offset: offset, Limit: limit, IDs: []string{}}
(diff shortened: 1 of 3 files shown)


dataset/dataset_test.go
```diff
@@ -847,6 +852,22 @@ func TestClient_GetOptions(t *testing.T) {
 			})
 		})
 
+		Convey("when GetOptions is called with a list of IDs containing an option with special characters", func() {
+			q := QueryParams{offset, limit, []string{"90+"}}
```
nshumoogum, line 856: For robustness, should we not be setting the keys to their values?

`q := QueryParams{Offset: offset, Limit: limit, Options: []string{"90+"}}`

Otherwise you are relying on the order which is ambiguous and can easily break when this is extended to cover more query parameters.

nshumoogum (changes requested):

dataset/dataset_test.go
DavidSubiros: Good point, I will also fix the rest of tests that do the same.

nshumoogum (approved):

PR #89, opened 2021-01-25
Add method that returns dataset current and next docs

jondewijones:
### What

Add method that returns dataset current and next docs

### Who can review

Anyone


justinpjose (approved):
Looks good. That was a nice PR to review

PR #90, opened 2021-01-25
Add error field to image api model

jessjenkins:
### What

Add an error field to the image model (similar to the one in the ImageDownload model). This means that when the image import/publish fail, a descriptive reason can be supplied as well. This requires a correlating change in `dp-image-api` (https://github.com/ONSdigital/dp-image-api/pull/47) 

### How to review

Ensure the new field has been added.

### Who can review

Anyone but me.


justinpjose (approved):
Looks good

PR #91, opened 2021-01-25
Feature/etag

DavidSubiros:
CAUTION! Breaking change

### What

- Changes in Filter API client to support Etag and If-Match:
  - Returns the ETag value for all calls that return it in a header ('write' and 'read' methods)
  - Requires If-Match for all calls that require it ('write' methods)

- Batch processing if-Match / ETag handling
  - Batch Getter fails if the returned ETag in a batch is different from the previous values.
  - Batch Processor offers a bool parameter to optionally check the ETag values from one batch to another. The latest ETag value is returned.
  - Batch Processor batch process function now receives the ETag value for the current batch

- Modified unit tests and mocks to verify ETags for all affected functions
- Created extra unit tests to validate new functionality for batch processing.

### How to review

- Make sure unit tests pass

### Who can review

Anyone


redhug1 (approved):
Looks OK

PR #105, opened 2021-03-30
Add method for getting batches of versions

jondewijones:
### What

Add method for getting batches of versions

### Who can review

Anyone

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,3 +4,11 @@ test:
 	go test -v -count=1 -race -cover ./...
 
 .PHONY: test
+
+audit:
+	go list -json -m all | nancy sleuth --exclude-vulnerability-file ./.nancy-ignore
+.PHONY: audit
+
+build:
+	go build ./...
+.PHONY: build
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ test:
 .PHONY: test
 
 audit:
-	go list -json -m all | nancy sleuth --exclude-vulnerability-file ./.nancy-ignore
+	go list -json -m all | nancy sleuth
 .PHONY: audit
 
 build:
--- a/dataset/data.go
+++ b/dataset/data.go
@@ -50,6 +50,7 @@ type List struct {
 	TotalCount int       `json:"total_count"`
 }
 
+// VersionList represents an object containing a list of datasets
 type VersionsList struct {
 	Items      []Version `json:"items"`
 	Count      int       `json:"count"`
--- a/dataset/data.go
+++ b/dataset/data.go
@@ -50,6 +50,14 @@ type List struct {
 	TotalCount int       `json:"total_count"`
 }
 
+type VersionsList struct {
+	Items      []Version `json:"items"`
+	Count      int       `json:"count"`
+	Offset     int       `json:"offset"`
+	Limit      int       `json:"limit"`
+	TotalCount int       `json:"total_count"`
+}
+
 // Version represents a version within a dataset
 type Version struct {
 	Alerts               *[]Alert             `json:"alerts"`
--- a/dataset/dataset.go
+++ b/dataset/dataset.go
@@ -499,7 +499,7 @@ func (c *Client) GetVersionsInBatches(ctx context.Context, userAuthToken, servic
 			versions.Count = b.TotalCount
 		}
 		for i := 0; i < len(b.Items); i++ {
-			if len(versions.Items) < i+b.Offset {
+			if len(versions.Items) < len(b.Items)+b.Offset {
 				return false, fmt.Errorf("versions.Items offset index out of bounds error. Expected length: %d, actual length: %d", i+b.Offset, len(versions.Items))
 			}
 			versions.Items[i+b.Offset] = b.Items[i]
--- a/dataset/dataset.go
+++ b/dataset/dataset.go
@@ -499,7 +499,7 @@ func (c *Client) GetVersionsInBatches(ctx context.Context, userAuthToken, servic
 			versions.Count = b.TotalCount
 		}
 		if len(versions.Items) < len(b.Items)+b.Offset {
-			return false, fmt.Errorf("versions.Items offset index out of bounds error. Expected length: %d, actual length: %d", i+b.Offset, len(versions.Items))
+			return false, fmt.Errorf("versions.Items offset index out of bounds error. Expected length: %d, actual length: %d", len(b.Items)+b.Offset, len(versions.Items))
 		}
 		for i := 0; i < len(b.Items); i++ {
 			versions.Items[i+b.Offset] = b.Items[i]
--- a/dataset/dataset.go
+++ b/dataset/dataset.go
@@ -62,6 +62,9 @@ type ErrInvalidDatasetAPIResponse struct {
 // DatasetsBatchProcessor is the type corresponding to a batch processing function for a dataset List.
 type DatasetsBatchProcessor func(List) (abort bool, err error)
 
+// VersionsBatchProcessor is the type corresponding to a batch processing function for a dataset List.
+type VersionsBatchProcessor func(VersionsList) (abort bool, err error)
+
 // OptionsBatchProcessor is the type corresponding to a batch processing function for dataset Options
 type OptionsBatchProcessor func(Options) (abort bool, err error)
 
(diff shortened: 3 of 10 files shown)


dataset/dataset.go
```diff
@@ -467,15 +476,63 @@ func (c *Client) GetVersions(ctx context.Context, userAuthToken, serviceAuthToke
...
+	//   - We do a single memory allocation for the final array, making the code more memory efficient.
+	var processBatch VersionsBatchProcessor = func(b VersionsList) (abort bool, err error) {
+		if len(versions.Items) == 0 { // first batch response being handled
+			versions.TotalCount = b.TotalCount
+			versions.Items = make([]Version, b.TotalCount)
+			versions.Count = b.TotalCount
+		}
+		for i := 0; i < len(b.Items); i++ {
+			versions.Items[i+b.Offset] = b.Items[i]
+		}
```
daiLlew, line 502: I think you need to be careful here and check that `i+b.Offset` is not > the length of items.

dataset/dataset.go
```diff
@@ -467,15 +476,63 @@ func (c *Client) GetVersions(ctx context.Context, userAuthToken, serviceAuthToke
...
+		return false, nil
```
daiLlew, line 503: I'm not sure I follow the purpose of returning the `abort bool` - it's always returning `false`. So my question is is this needed?

dataset/dataset.go
```diff
@@ -467,15 +476,63 @@ func (c *Client) GetVersions(ctx context.Context, userAuthToken, serviceAuthToke
...
+	}
+
+	// call dataset API GetOptions in batches and aggregate the responses
+	if err := c.GetVersionsBatchProcess(ctx, userAuthToken, serviceAuthToken, downloadServiceAuthToken, collectionID, datasetID, edition, processBatch, batchSize, maxWorkers); err != nil {
+		return VersionsList{}, err
```
daiLlew, line 508: Is there a reason why you need to create a new `VersionList{}` here? This method uses a named return parameter so can you return `versions` instead? I think if we are going to use named return values then we should return those params in all cases - otherwise it's misleading and potentially confusing.

dataset/dataset.go
```diff
@@ -467,15 +476,63 @@ func (c *Client) GetVersions(ctx context.Context, userAuthToken, serviceAuthToke
...
+	}
+
+	return versions, nil
+}
+
+// GetVersionsBatchProcess gets the datasets from the dataset API in batches, calling the provided function for each batch.
+func (c *Client) GetVersionsBatchProcess(ctx context.Context, userAuthToken, serviceAuthToken, downloadServiceAuthToken, collectionID, datasetID, edition string, processBatch VersionsBatchProcessor, batchSize, maxWorkers int) error {
+
+	// for each batch, obtain the dimensions starting at the provided offset, with a batch size limit,
+	// or the subste of IDs according to the provided offset, if a list of optionIDs was provided
```
daiLlew, line 518: typo - subset.

dataset/dataset.go
```diff
@@ -467,15 +476,63 @@ func (c *Client) GetVersions(ctx context.Context, userAuthToken, serviceAuthToke
...
+	batchGetter := func(offset int) (interface{}, int, string, error) {
+		b, err := c.GetVersions(ctx, userAuthToken, serviceAuthToken, downloadServiceAuthToken, collectionID, datasetID, edition, &QueryParams{Offset: offset, Limit: batchSize})
+		return b, b.TotalCount, "", err
+	}
+
+	// cast and process the batch according to the provided method
+	batchProcessor := func(b interface{}, batchETag string) (abort bool, err error) {
+		v, ok := b.(VersionsList)
+		if !ok {
+			return true, errors.New("wrong type")
```
daiLlew, line 528: I think this error could have more context to make it clearer specifically what the problem is e.g.:

`"version batch processor error wrong type received expected VersionList but was X"`

How to get the type of an interface - https://play.golang.org/p/dNY92SqVfvs

dataset/dataset.go
```diff
@@ -446,8 +449,14 @@ func (c *Client) GetEditions(ctx context.Context, userAuthToken, serviceAuthToke
 }
 
 // GetVersions gets all versions for an edition from the dataset api
-func (c *Client) GetVersions(ctx context.Context, userAuthToken, serviceAuthToken, downloadServiceAuthToken, collectionID, datasetID, edition string) (m []Version, err error) {
+func (c *Client) GetVersions(ctx context.Context, userAuthToken, serviceAuthToken, downloadServiceAuthToken, collectionID, datasetID, edition string, q *QueryParams) (m VersionsList, err error) {
 	uri := fmt.Sprintf("%s/datasets/%s/editions/%s/versions", c.hcCli.URL, datasetID, edition)
+	if q != nil {
+		if err := q.Validate(); err != nil {
+			return VersionsList{}, err
```
daiLlew, line 456: This method uses named return vars so for consistency and to avoid confusion you should return the named vars. A function that declares named return vars implies the func will return those specific vars so it's misleading and potentially confusing to return other vars. I'm aware this is picky but I think its an important point to adhere to when using a this feature.

Update  `if err := q.Validate...`  to  `if err = q.Validate...`  i.e. Assign the err returned from the function to the existing ` err` named parameter.

Then simply 
```
return
```

daiLlew (changes requested):
You've added new functionality to this package so I think we need to also write tests to cover the new functionality - happy and unhappy scenarios.

dataset/dataset.go
daiLlew, line 502: Is there a reason why this needs to be an inline anonymous func? Could it be pulled put into stand function? It would be less code in 1 function and also reusable.

daiLlew (changes requested):

dataset/data.go
```diff
@@ -50,6 +50,14 @@ type List struct {
 	TotalCount int       `json:"total_count"`
 }
 
+type VersionsList struct {
```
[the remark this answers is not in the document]
jondewijones: 👍

dataset/dataset.go
jondewijones: 🤷 was c&p from other batch getters/processors

dataset/dataset.go
jondewijones: 🤷 was c&p from other batch getters/processors

dataset/dataset.go
jondewijones: Changed. Was same as other batch getters/processors

dataset/dataset.go
jondewijones: 👍

dataset/dataset.go
jondewijones: 👍

dataset/dataset.go
jondewijones: 👍

dataset/dataset.go
jondewijones: 🤷 was c&p from other batch getters/processors

Makefile
```diff
@@ -4,3 +4,11 @@ test:
 	go test -v -count=1 -race -cover ./...
 
 .PHONY: test
+
+audit:
+	go list -json -m all | nancy sleuth --exclude-vulnerability-file ./.nancy-ignore
```
jessjenkins, line 9: No need to specify the exclude file, `.nancy-ignore` is already the default anyway so this flag does nothing.
```suggestion
	go list -json -m all | nancy sleuth
```

dataset/dataset.go
daiLlew: I still think you need to be careful as you are making assumption about the length of items, if that assumption is wrong this will fail and be a defect. I don't think it too much additional effort to check and error if its not as expected?

dataset/dataset.go
daiLlew: I realised after post this comment this function definition is a pattern used by the batch processor so it's fine as is because this is the expected structure for an established existing function.

dataset/dataset.go
```diff
@@ -467,15 +477,68 @@ func (c *Client) GetVersions(ctx context.Context, userAuthToken, serviceAuthToke
...
+	//   - We can guarantee, even with concurrent calls, that values are returned in the same order that the API defines, by offsetting the index.
+	//   - We do a single memory allocation for the final array, making the code more memory efficient.
+	var processBatch VersionsBatchProcessor = func(b VersionsList) (abort bool, err error) {
+		if len(versions.Items) == 0 { // first batch response being handled
+			versions.TotalCount = b.TotalCount
+			versions.Items = make([]Version, b.TotalCount)
+			versions.Count = b.TotalCount
+		}
+		for i := 0; i < len(b.Items); i++ {
+			if len(versions.Items) < i+b.Offset {
```
daiLlew, line 502: Sorry I should have spotted this during the chat but if you update this to:

```
if len(versions.Items) < len(b.Items) + b.Offset {
    return false, fmt.Errorf(....)
}
```
You can put it before the loop and it will be more efficient than checking in each iteration.

daiLlew (approved):

PR #108, opened 2021-04-27
added order property to hierarchy model, so that it is correclty mapp…

DavidSubiros:
### What

Added order property to hierarchy model, so that it's mapped from the response returned by Hierarchy API.

### How to review

- Make sure unit tests pass

### Who can review

anyone

### DIFF
--- a/hierarchy/hierarchy_test.go
+++ b/hierarchy/hierarchy_test.go
@@ -185,90 +185,138 @@ func getMockHierarchyAPI(expectRequest http.Request, mockedHTTPResponse MockedHT
 func TestClient_GetRoot(t *testing.T) {
 	instanceID := "foo"
 	name := "bar"
-	model := `{"label":"","has_data":true}`
-	Convey("When bad request is returned", t, func() {
+
+	Convey("Given a bad request API response, then the expected error is returned", t, func() {
 		ts, mockedAPI := getMockHierarchyAPI(http.Request{Method: "GET"}, MockedHTTPResponse{StatusCode: 400, Body: ""})
 		_, err := mockedAPI.GetRoot(ctx, instanceID, name)
-		So(err, ShouldNotBeNil)
+		So(err, ShouldResemble, &ErrInvalidHierarchyAPIResponse{
+			actualCode:   400,
+			expectedCode: 200,
+			uri:          "/hierarchies/foo/bar",
+		})
 		ts.Close()
 	})
 
-	Convey("When server error is returned", t, func() {
+	Convey("Given a server error API response, then the expected error is returned", t, func() {
 		ts, mockedAPI := getMockHierarchyAPI(http.Request{Method: "GET"}, MockedHTTPResponse{StatusCode: 500, Body: "qux"})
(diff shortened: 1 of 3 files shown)


hierarchy/hierarchy_test.go
```diff
@@ -185,90 +185,138 @@ func getMockHierarchyAPI(expectRequest http.Request, mockedHTTPResponse MockedHT
...
 		ts.Close()
 	})
 
-	Convey("When server error is returned", t, func() {
+	Convey("Given a server error API response, then the expected error is returned", t, func() {
 		ts, mockedAPI := getMockHierarchyAPI(http.Request{Method: "GET"}, MockedHTTPResponse{StatusCode: 500, Body: "qux"})
 		mockedAPI.hcCli.Client.SetMaxRetries(2)
 		_, err := mockedAPI.GetRoot(ctx, instanceID, name)
-		So(err, ShouldNotBeNil)
+		So(err, ShouldResemble, &ErrInvalidHierarchyAPIResponse{
```
rav-pradhan, line 204: Could it be worth extracting this into a helper function that builds the expected response for us? In case the expected response ever changes in the future (e.g. additional field added). We can then just pass the new argument instead

hierarchy/hierarchy_test.go
```diff
@@ -185,90 +185,138 @@ func getMockHierarchyAPI(expectRequest http.Request, mockedHTTPResponse MockedHT
...
 	name := "bar"
-	model := `{"label":"","has_data":true}`
-	Convey("When bad request is returned", t, func() {
+
+	Convey("Given a bad request API response, then the expected error is returned", t, func() {
 		ts, mockedAPI := getMockHierarchyAPI(http.Request{Method: "GET"}, MockedHTTPResponse{StatusCode: 400, Body: ""})
 		_, err := mockedAPI.GetRoot(ctx, instanceID, name)
-		So(err, ShouldNotBeNil)
+		So(err, ShouldResemble, &ErrInvalidHierarchyAPIResponse{
+			actualCode:   400,
```
rav-pradhan, line 193: Perhaps throughout the tests we could use the Go consts for the http codes? E.g. `http.StatusBadRequest` here?

rav-pradhan (commented):
Looks good - just a couple small potential refactors we could make. Let me know if you have any thoughts

hierarchy/hierarchy_test.go
DavidSubiros: Yes, good point. I can do the same for all the status codes

hierarchy/hierarchy_test.go
DavidSubiros: There is a method to create the error struct, maybe we can use it:
```
func NewErrInvalidHierarchyAPIResponse(expectedCode, actualCode int, uri string) error {
```

DavidSubiros: I used this method to generate the expected responses, but also added another test to test this method itself.

rav-pradhan (approved):

PR #112, opened 2021-06-08
Added recipe API client package, with GET /recipes/{id} call

DavidSubiros:
### What

The new Cantabular import pipeline requires some calls to recipe API by different servies, hence we need a client in dp-api-clients-go for it. We can also modify the existing import API to use this client after it is merged.

- Created `recipe` package
- Created constructors and errors consistent with other packages
- Created GetRecipe func, which wraps `GET /recipes/{id}` call
- The main part of the code has been migraged from `dp-import-cantabular-dataset` repo (which will use this client instead, once merged)
- Added unit tests

### How to review

- Make sure unit tests pass

### Who can review

Anyone


recipe/recipe.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	if err := resp.Body.Close(); err != nil {
+		log.Event(ctx, "error closing http response body", log.ERROR, log.Error(err))
+	}
+}
+
+// GetRecipe from an ID
+func (c *Client) GetRecipe(ctx context.Context, userAuthToken, serviceAuthToken, recipeID string) (*Recipe, error) {
+	uri := fmt.Sprintf("%s/recipes/%s", c.hcCli.URL, recipeID)
+	clientlog.Do(ctx, "retrieving recipe", service, uri)
```
ChrysmOre, line 58: There was a mention in the guild meeting yesterday that perhaps we shouldn't be logging at all in any of our libraries (I removed the logging I had because of this) and let the services handle the logging.. this sounded like a good way to me, not sure what you think? It does look that logging package was specifically designed for client though..

recipe/recipe.go
```diff
@@ -0,0 +1,109 @@
...
+
+	resp, err := c.doGetWithAuthHeaders(ctx, userAuthToken, serviceAuthToken, uri)
+	if err != nil {
+		return nil, fmt.Errorf("failed to call recipe api: %w", err)
+	}
+
+	defer CloseResponseBody(ctx, resp)
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, &Error{
```
ChrysmOre, line 68: It seems you removed the errorResponse func that handles the error response from the recipe api.. Is it because the recipe api never returns a response body with an error?

recipe/recipe.go
DavidSubiros: Yes, I added it for consistency with other services. I can remove it, and maybe at some point, we remove the logging from all other packages in the library.

recipe/recipe.go
DavidSubiros: Thanks, I will add it back - I originally used another Error type here, which didn't have a response body.

recipe/recipe.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+}
+
+// Checker calls recipe api health endpoint and returns a check object to the caller.
+func (c *Client) Checker(ctx context.Context, check *health.CheckState) error {
+	return c.hcCli.Checker(ctx, check)
+}
+
+// CloseResponseBody closes the response body and logs an error if unsuccessful
+func CloseResponseBody(ctx context.Context, resp *http.Response) {
```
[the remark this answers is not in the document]
DavidSubiros: I will update all clients, for consistency

recipe/recipe.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, fmt.Errorf("failed to call recipe api: %w", err)
+	}
+
+	defer CloseResponseBody(ctx, resp)
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, &Error{
+			err:        errors.New("wrong status code, expected 200 OK"),
+			statusCode: resp.StatusCode,
+			logData:    log.Data{},
```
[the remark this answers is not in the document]
DavidSubiros: Ok, I will remove it.

ChrysmOre (approved):

PR #114, opened 2021-06-09
Add Cantabular Client

ChrysmOre:
### What

Added a client package for interacting with Cantabular API

### How to review

- Check unit tests pass and make sense.. 
- f you're feeling adventurous write a small program that uses the client to interact with Cantabular and experiment with some queries.
- I'm not actually 100% sure this is the right repository to put it in, but it seems to make the most sense. It will be used by all of the services that interact with Cantabular.

### Who can review

Anyone


cantabular/codebook_test.go
```diff
@@ -0,0 +1,136 @@
...
+		cantabularClient := cantabular.NewClient(
+			mockHttpClient,
+			cantabular.Config{},
+		)
+
+		Convey("When the GetCodebook method is called", func() {
+			req := cantabular.GetCodebookRequest{}
+			cb, err := cantabularClient.GetCodebook(testCtx, req)
+
+			Convey("Then the status code 500 should be recoverable from the error", func() {
```
ChrysmOre, line 68: Should this actually be a 502?

DavidSubiros: I think that 500 is OK, as this is just a library talking to Cantabular service. Then the caller of this library should decide what to do. If the caller is an API, then it can return 502 in this scenario, so the caller of the API knows that some service down the line is failing (not the API directly). If the caller is a Kafka service, then it might decide to do something different.

DavidSubiros (approved):
I really like the approach to error handling, and I think it would be a good idea to do the same for other clients.

In the future we may want to wrap a health check client, similar to other clients, but as discussed, we are not aware of any `/health` endpoint in cantabular, so this may be something to consider in the future (not in the scope of this PR)

PR #117, opened 2021-06-14
Update documentation link to health package to be relative link

nshumoogum:
### What

Previously being an absolute link - it would fail due to branch being deleted.

Now the absolute link will always work (unless the file is removed) on any branch. It will also prevent the user when clicking the link to jump to a different branch (which would happen if it were an absolute link).

### How to review

Check health link works

### Who can review

Anyone


dandwal (approved):
Link works

PR #118, opened 2021-06-14
implemented PostInstance, requried by import API

DavidSubiros:
### What

- Implemented `PostInstance` method for dataset API client.
- Created unit tests

This is required by import API, which currently uses an implementation in the service itself and is being replaced with this one.

### How to review

- Make sure new tests make sense
- Make sure unit tests pass

### Who can review

Anyone

### DIFF
--- a/dataset/data.go
+++ b/dataset/data.go
@@ -59,6 +59,35 @@ type VersionsList struct {
 	TotalCount int       `json:"total_count"`
 }
 
+// Instance which presents a single dataset being imported
+type NewInstance struct {
+	InstanceID        string               `json:"id,omitempty"`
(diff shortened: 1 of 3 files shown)


dataset/data.go
```diff
@@ -59,6 +59,35 @@ type VersionsList struct {
 	TotalCount int       `json:"total_count"`
 }
 
+// Instance which presents a single dataset being imported
```
DavidSubiros, line 62: Note: these models have been copied from import API:
https://github.com/ONSdigital/dp-import-api/blob/develop/models/models.go#L115

CarlHembrough (approved):

PR #119, opened 2021-06-15
Add url to logData in several places

ChrysmOre:
### What

Added some addition logData to errors in Cantabular API as I found it useful in debugging.

### How to review

Check makes sense, unit tests pass

### Who can review

anyone


cantabular/codebook.go
```diff
@@ -46,7 +48,7 @@ func (c *Client) GetCodebook(ctx context.Context, req GetCodebookRequest) (*GetC
 			fmt.Errorf("failed to read response body: %s", err),
 			res.StatusCode,
 			log.Data{
-				"response_body": string(b),
+				"url": url,
```
DavidSubiros, line 51: makes sense, if we can't ready the response Body, maybe the URL is more useful.

DavidSubiros (approved):
lgtm

PR #120, opened 2021-06-16
Update client and data structures to use Cantabular v9

ChrysmOre:
### What

Updated Cantabular Client to use v9 of the software
- Updated urls to use /v9
- Updated data structures to use new MapFrom form

### How to review

- Check tests pass, code makes sense
- Use client to make a request against Cantabular and check all data is returned correctly

### Who can review

Anyone


franmoore05 (approved):
LGTM

PR #121, opened 2021-06-18
fix download files in datasets that have absolute url

s-xton:
### What

Fixed an issue where an absolute url for a download file in data.json was still appended onto the page url: 

### How to review

Check against existing content. If you are reviewing pls contact me for a URL with an absolute path

### Who can review

Anyone with sufficient knowledge of this repo

### DIFF
--- a/ci/lint.yml
+++ b/ci/lint.yml
@@ -1,0 +1,15 @@
+---
+
+platform: linux
+
+image_resource:
+  type: docker-image
+  source:
+    repository: golang
+    tag: 1.15.7
+
+inputs:
+  - name: dp-api-clients-go
+
+run:
+  path: dp-api-clients-go/ci/scripts/lint.sh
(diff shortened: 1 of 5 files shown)


ChrysmOre (dismissed):

nshumoogum (dismissed):
Nice spot

ci/lint.yml
```diff
@@ -0,0 +1,15 @@
...
+platform: linux
+
+image_resource:
+  type: docker-image
+  source:
+    repository: golang
+    tag: 1.15.7
+
+inputs:
+  - name: dp-dataset-api
```
nshumoogum, line 12: I believe this be `dp-api-clients-go`

nshumoogum (changes requested):

ci/lint.yml
nshumoogum: Also wondering whether it is worth upgrading library to use latest go version - i think it is now on 1.16.5

nshumoogum (approved):

PR #125, opened 2021-06-30
Replacement for setUserAuthToken introduced

dandwal:
### What

As part of the new Authentication work, we will be replacing X-Florence-Token which is validated via Zebedee with a new token that uses the Authorization header instead which is validated by the dp-identity-api.

But wait we need a smooth transition! Yes, and so we have a new function for setting authentication header `SetAuthToken` this will set both the old Zebedee token and the Florence token. There is work in progress to make requests to the dp-identity-api that fail auth to then check Zebedee with the X-Florence-Token. This work is WIP.

You may notice that the serviceAuthToken also uses the header `Authorization`! Won't there be a collision? - Yes. The idea behind this work is to handle these collisions later, potentially what we will do is check the dp-identity-api to see if it is a user token, if not then check Zebedee to see if it is a service token. This is an interim solution though.

All places that referred to SetUserAuthToken have been replaced with SetAuthToken

Why SetAuthToken instead of SetUserAuthToken, in the distant future service auth tokens will be removed from Zebedee like we are doing currently with user tokens.

### How to review

Check the code

### Who can review

Anyone except me

### DIFF
--- a/codelist/codelist.go
+++ b/codelist/codelist.go
@@ -297,7 +297,7 @@ func (c *Client) doGetWithAuthHeaders(ctx context.Context, userAuthToken string,
 }
 
 func setAuthenticationHeaders(req *http.Request, userAuthToken, serviceAuthToken string) error {
-	err := headers.SetUserAuthToken(req, userAuthToken)
+	err := headers.SetAuthToken(req, userAuthToken)
 	if err != nil && err != headers.ErrValueEmpty {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


daiLlew (changes requested):
I've posted a question in slack about this.  Can we have a team chat to make sure we're all on the same page and make sure we're fully aware of the impacts?

I've 'requested changes` as I am worried this could have unintended consequences.

janderson2:
I believe this should be ok so long as all usages of Set always set the client first and allow the service token to overwrite (this is the current behaviour within this repo, but we should share with the team so that any repos that pull in this update will know to check the order).

In terms of auth itself. The codebase always takes the user over the service auth by default so our change will be backwards compatible there for now: https://github.com/ONSdigital/dp-api-clients-go/blob/main/identity/authentication.go#L75. We will definitely need to replace this auth implementation later and this is planned work.

daiLlew (approved):

PR #128, opened 2021-07-07
Upgrade module to v2

dandwal:
### What

In a previous release there was a breaking change: https://github.com/ONSdigital/dp-api-clients-go/pull/125 this was going to be tagged as a 2.0.0 release, however, Go doesn't recognise a v1 to v2 change this simply. This PR fixes this issue and updates the required go.mod file and references to make this library compatible with a v2 change.

**Release 2.0.0** 🥳 

- Upgrade go to 1.16
- Upgrade dependencies

### How to review

Check that the version upgrade has been done correctly for comparison check out our other /v2 libraries:
- [dp-kafka](https://github.com/ONSdigital/dp-kafka/blob/main/go.mod)
- [dp-log](https://github.com/ONSdigital/log.go/blob/master/go.mod)

### Who can review

Anyone except me


go.mod
```diff
@@ -1,14 +1,14 @@
-module github.com/ONSdigital/dp-api-clients-go
+module github.com/ONSdigital/dp-api-clients-go/v2
```
daiLlew, line 1: I'm not sure this is right. I think this will rename the module which would break it for everything using it and would also make the original v1 iterations unavailable.

I think the modules approach is to create a new `v2` package within the lib and then put the new v2 functionality in there. Users then update their import to use `/v2`.

dandwal: I can deffo do this approach if needed, however, this hasn't been the way DP has been going to v2 of packages in the past (examples in PR, dp-kafka and dp-log). I think the idea is that we don't support previous versions of a package and people should bump their dependencies to the latest major version

daiLlew: There is an official Go modules guide here which outlines the standard  and how to solves this problem I just want to check we are following this pattern, I think we are mostly but not completely.

https://blog.golang.org/v2-go-modules

dandwal:
We will ignore the previous dp upgraded method and use the GO standard

PR #131, opened 2021-07-08
Release v2 of the go mod

dandwal:
### What

Version 2 go mod 
As per discussions on Slack. Main branch will contain the V2 module content.
A separate branch has been taken for v1 and currently exists. All v1 tags will still work and that branch can be edited for any v1 changes.

### Who can review

Anyone except me


HeJones (approved):
It looks good to me and from reading the thread in slack I think it aligns with what was agreed

PR #135, opened 2021-07-27
add type, isBasedOn, variable and numberOfOptions

ChrysmOre:
### What

Added new Cantabular model fields to dataset client to match dp-dataset-api models

### How to review

Make sure code makes sense, tests pass.
Can check relative values in dp-dataset-api/models/ (type Instance and type Dimension) to see they match in type and JSON tags

### Who can review

Anyone

### DIFF
--- a/dataset/data.go
+++ b/dataset/data.go
@@ -125,16 +125,20 @@ type UpdateInstance struct {
 	NumberOfObservations int64                `json:"total_observations,omitempty"`
 	ImportTasks          *InstanceImportTasks `json:"import_tasks,omitempty"`
 	CSVHeader            []string             `json:"headers,omitempty"`
+	Type                 string               `json:"type,omitempty"`
+	IsBasedOn            *IsBasedOn           `json:"is_based_on,omitempty"`
 }
 
 // VersionDimension represents a dimension model nested in the Version model
 type VersionDimension struct {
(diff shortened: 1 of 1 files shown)


daiLlew (commented):
Code looks good to me but I do have a general question:

You may have already considered this and have a plan but Ill ask anyway - have we considered how backward compatibility will work for existing datasets / instances? 

If we are adding new fields to these models will that affect existing data stored in Mongo and the graph dbs? Do we need to update existing data or use a default for these new fields?

ChrysmOre:
> Code looks good to me but I do have a general question:
> 
> You may have already considered this and have a plan but Ill ask anyway - have we considered how backward compatibility will work for existing datasets / instances?
> 
> If we are adding new fields to these models will that affect existing data stored in Mongo and the graph dbs? Do we need to update existing data or use a default for these new fields?

The `omitempty` field keeps these changes backwards compatible. The related fields in Mongo also have the `omitempty` bson tag, so existing data should pass through unaffected. The tests in `dp-dataset-api` confirm this. Additionally these changes are to the `v2` of dp-api-clients go which is still in beta, but the directly affecting changes have already been made in `dp-dataset-api` few weeks ago.

redhug1 (approved):
LGTM

PR #137, opened 2021-07-28
Implement cantabular healthcheck and add 'type' in dataset Instance model

DavidSubiros:
### What

- Implement cantabular healthcheck
- Add `type` to dataset instance model

### How to review

- Make sure unit tests pass
- (note) I tried to use the changes in the import process and I can see that the instance type is correctly used and the cantabular health check is correctly assessed.

### Who can review

anyone


redhug1 (approved):
Code changes look sensible.
Tests Pass.
LGTM

PR #138, opened 2021-08-09
Add method to put download info to instances (dataset api)

DavidSubiros:
### What

Improved the Downloads model in UpdateInstance (according to models defined in dp-dataset-api)

### How to review

- Make sure code changes make sense. You may check the dataset api models [here](https://github.com/ONSdigital/dp-dataset-api/blob/develop/models/instance.go#L17)

### Who can review

anyone


redhug1 (approved):
Code changes look reasonable

PR #139, opened 2021-08-10
add aroundONS property to zebedee client model

AngelaLW:
### What

add around ONS property to zebedee client model

### How to review

review code and test changes

### Who can review

anyone

### DIFF
--- a/zebedee/data.go
+++ b/zebedee/data.go
@@ -139,6 +139,7 @@ type TimeseriesDescription struct {
 type HomepageContent struct {
 	Intro           Intro               `json:"intro"`
 	FeaturedContent []Featured          `json:"featuredContent"`
+	AroundONS 		[]Featured          `json:"aroundONS"`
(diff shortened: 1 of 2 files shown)


zebedee/data.go
```diff
@@ -139,6 +139,7 @@ type TimeseriesDescription struct {
 type HomepageContent struct {
 	Intro           Intro               `json:"intro"`
 	FeaturedContent []Featured          `json:"featuredContent"`
+	AroundONS 		[]Featured          `json:"aroundONS"`
```
daiLlew, line 142: Question: Do we need to handle backward compatibility here? I.e. Zebeded does not have the model? I think it should be fine we should test the client against a Zebedee version that does not have this new field to make sure it doesn't break.
If it does we could make it optional

AngelaLW: I've given it a test locally with zebedee not having the model property and the client is functioning fine, the rest of the page displays ok.

daiLlew (approved):

PR #141, opened 2021-08-11
Feature/update ctblr client for ext api queries

ChrysmOre:
### What

For Cantabular Client:

- Added GraphQL Query for Static Dataset, new data structures and unit tests
- Includes a breaking change to the NewClient() func signature (change will need to be versioned)

### How to review

- Check tests pass and code makes. 
- Not much can be tested manually without the service/s that use the imported client.
- Code change will be breaking to existing services importing Cantabular Client.

### Who can review

Anyone


cantabular/client.go
```diff
@@ -13,23 +13,40 @@ import (
...
+	host       string
+	extApiHost string
 }
 
 // NewClient returns a new Client
-func NewClient(ua httpClient, cfg Config) *Client {
-	return &Client{
-		ua:   ua,
-		host: cfg.Host,
+func NewClient(cfg Config, ua httpClient, g GraphQLClient) *Client {
```
DavidSubiros, line 32: Just a suggestion, maybe we can create extra constructors?
e.g. 
- `NewClient(cfg Config) *Client` : creates an httpClient and a GraphQLClient
- `NewClientWithHttp(cfg Config, ua httpClient) *Client` : uses the provided httpClient and creates a new GraphQLClient
- `NewClientWithHttpAndGrapQL(cfg Config, ua httpClient, g GraphQLClient) *Client` : uses the provided httpClient and GraphQLClient

Then we could also create getters for the httpClient and the GraphQLCLient, so that they can be reused by the caller if required

Do you think it would be a good approach?

DavidSubiros (approved):
Approved, but I left a suggestion of possibly having multiple constructors?

cantabular/client.go
ChrysmOre: Yeah I went back and forth with how best to do this. Perhaps it couldn't hurt to include extra constructors. I did want to include getters. It will be breaking either way so I'm to go ahead as is and maybe add more a little down the line

PR #143, opened 2021-08-19
add parsing for graphql error

ChrysmOre:
### What

Added parsing for GraphQL errors returned from Cantabular. GraphQL errors don't come back as actual
errors or with an error code, rather you have to check whether the error field in the response has been populated. 
With this change the Cantabular Client will now convert this into an error with the appropriate status code.

### How to review

Check code changes make sense, tests pass.

To test locally you can cause a GraphQL error by modifying the recipe to have too many variables or editing the code to pass incorrect variables.

### Who can review

Anyone


DavidSubiros (approved):
LGTM

PR #147, opened 2021-08-31
integrate table parsing logic for Cantabular client

ChrysmOre:
### What

Have ported the response parsing logic and tests from dp-cantabular-csv-exporter so it can be reused in dp-observation-api (and anywhere else that needs it in the future)

### How to review

Check tests pass and code makes sense, only expected code is present.

### Who can review

Anyone

### DIFF
--- a/cantabular/table.go
+++ b/cantabular/table.go
@@ -1,9 +1,80 @@
 package cantabular
 
+import (
+	"bufio"
+	"bytes"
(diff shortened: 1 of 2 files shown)


DavidSubiros (approved):
LGTM - The parsing logic is the same as in Cantabular csv exporter, and the unit tests are successful.

PR #157, opened 2021-11-12
improved Stream error handling, closing and var names

DavidSubiros:
### What

- Renamed r/w to pipeReader/pipeWriter to improve code readability
- Transformer func calls `CloseWithError` if the transform func returned an error (and `Close` otherwise)
- Fixed tests accordingly: we no longer see the `EOF` error, and any error in the transformer is passed to the consmer.

- Pass the context provided to `StaticDatasetQueryStreamCSV` down to the json streamer iterator, and abort processing if the context is cancelled 

Use XLSX exporter as a reference.

### How to review

- Make sure unit tests pass

### Who can review

Anyone

### DIFF
--- a/cantabular/dimension.go
+++ b/cantabular/dimension.go
@@ -1,6 +1,10 @@
 package cantabular
 
-import "errors"
+import (
+	"context"
+	"errors"
+	"fmt"
+)
 
 type (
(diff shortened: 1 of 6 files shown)


redhug1 (approved):
LGTM

PR #159, opened 2021-11-25
created iota enum for import api states

DavidSubiros:
### What

As part of updating dependencies for dp-cantabular- (trello card: https://trello.com/c/NxIax7FQ/5349-start-stop-consuming-on-healthy-unhealthy-for-dp-import-cantabular-dataset-8), I found a dependency with `dp-import-api`, which was only required in order to read the 'failed' state constant.

Following the same strategy that we implemented in `dataset api` client, in this PR we are creating an iota enum that matches the import api state constants.

Once this is merged, `dp-import-cantabular-dataset` will no longer require the dependency with dp-import-api go module.

### How to review

- Make sure code change make sense
- Make sure unit tests changes make sense and cover the iota and its usage
- Make sure unit tests pass

### Who can review

anyone

### DIFF
--- a/importapi/importapi.go
+++ b/importapi/importapi.go
@@ -18,6 +18,35 @@ import (
 
 const service = "import-api"
 
+// State - iota enum of possible states
+type State int
+
+// stateData represents a json with a single state filed
+type stateData struct {
+	State string `json:"state"`
+}
+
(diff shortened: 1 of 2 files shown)


redhug1 (approved):
LGTM

PR #160, opened 2021-11-26
Add published index endpoint for search reindexing

jessjenkins:
### What

Added PublishedIndex endpoint to zebedee client
- Includes paging parameters (although zebedee currently ignores these)
- Corresponding zebedee PR: https://github.com/ONSdigital/zebedee/pull/551

### How to review

Check code and unit tests

### Who can review

Anyone but me

### DIFF
--- a/zebedee/client.go
+++ b/zebedee/client.go
@@ -469,6 +469,31 @@ func (c *Client) GetPublishedData(ctx context.Context, uriString string) ([]byte
 	return content, nil
 }
 
+// GetPublishedIndex returns PublishedIndex
+func (c *Client) GetPublishedIndex(ctx context.Context, params *PublishedIndexRequestParams) (PublishedIndex, error) {
+	reqURL := "/publishedindex"
+
+	if params != nil {
+		reqURL = fmt.Sprintf("%s?offset=%d",reqURL, params.Offset)
+		if params.Limit != nil {
+			reqURL = fmt.Sprintf("%s&limit=%d",reqURL, *(params.Limit))
+		}
+	}
+
+	b, _, err := c.get(ctx, "", reqURL)
+
+	if err != nil {
+		return PublishedIndex{}, err
+	}
+
+	var publishedIndex PublishedIndex
+	if err = json.Unmarshal(b, &publishedIndex); err != nil {
+		return publishedIndex, err
+	}
+
+	return publishedIndex, nil
+}
+
 // closeResponseBody closes the response body and logs an error if unsuccessful
 func closeResponseBody(ctx context.Context, resp *http.Response) {
 	if resp.Body != nil {
--- a/zebedee/client.go
+++ b/zebedee/client.go
@@ -474,14 +474,13 @@ func (c *Client) GetPublishedIndex(ctx context.Context, params *PublishedIndexRe
 	reqURL := "/publishedindex"
 
 	if params != nil {
-		reqURL = fmt.Sprintf("%s?offset=%d",reqURL, params.Offset)
+		reqURL = fmt.Sprintf("%s?offset=%d", reqURL, params.Offset)
 		if params.Limit != nil {
-			reqURL = fmt.Sprintf("%s&limit=%d",reqURL, *(params.Limit))
+			reqURL = fmt.Sprintf("%s&limit=%d", reqURL, *(params.Limit))
 		}
 	}
 
 	b, _, err := c.get(ctx, "", reqURL)
-
 	if err != nil {
 		return PublishedIndex{}, err
 	}
--- a/zebedee/client_test.go
+++ b/zebedee/client_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
@@ -542,7 +543,7 @@ func TestClient_PublishedDataEndpoint(t *testing.T) {
 		httpClient := newMockHTTPClient(response, nil)
 		zebedeeClient := newZebedeeClient(httpClient)
 
-		Convey("when zebedeeClient.GetPublised is called", func() {
+		Convey("when zebedeeClient.GetPublishedData is called", func() {
 			testContent, err := zebedeeClient.GetPublishedData(ctx, testURIString)
 
 			Convey("then the expected content is returned", func() {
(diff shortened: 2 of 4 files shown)


goofballLogic:
Do we have standards to review against? Or is it just "look for anything" ?

zebedee/client.go
```diff
@@ -469,6 +469,31 @@ func (c *Client) GetPublishedData(ctx context.Context, uriString string) ([]byte
...
+
+	if params != nil {
+		reqURL = fmt.Sprintf("%s?offset=%d",reqURL, params.Offset)
+		if params.Limit != nil {
+			reqURL = fmt.Sprintf("%s&limit=%d",reqURL, *(params.Limit))
+		}
+	}
+
+	b, _, err := c.get(ctx, "", reqURL)
+
```
rahulPalliyalil, line 484: (suggestion): Extra space, I think the space is not needed from func return and err check

zebedee/client_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
```
rahulPalliyalil, line 7: (fyi): just note that ioutil is deprecated from go 1.17 and is moved to the io package. So an update of the go version requires updating this to io.

zebedee/client.go
```diff
@@ -469,6 +469,31 @@ func (c *Client) GetPublishedData(ctx context.Context, uriString string) ([]byte
 	return content, nil
 }
 
+// GetPublishedIndex returns PublishedIndex
+func (c *Client) GetPublishedIndex(ctx context.Context, params *PublishedIndexRequestParams) (PublishedIndex, error) {
+	reqURL := "/publishedindex"
```
goofballLogic, line 474: I notice in the functions above, they form reqURL using `c.createRequestURL` whereas this one is hard coded

zebedee/client_test.go
```diff
@@ -607,3 +608,95 @@ func TestClient_PublishedDataEndpoint(t *testing.T) {
 		})
 	})
 }
+
+func TestClient_PublishedIndexEndpoint(t *testing.T) {
+	ctx := context.Background()
```
rahulPalliyalil, line 613: (minor): Can these tests be run in parallel ? t.Parallel()

rahulPalliyalil (dismissed):

zebedee/client.go
jessjenkins: Other functions have dynamic parameters and can take a collectionId. This one only has a simple endpoint so this isn't currently needed and can simply be a string.

zebedee/client_test.go
jessjenkins: Thanks. I'll replace with `os.ReadFile(…)` as per deprecation notes in https://pkg.go.dev/io/ioutil#ReadFile

zebedee/client_test.go
```diff
@@ -607,3 +608,95 @@ func TestClient_PublishedDataEndpoint(t *testing.T) {
...
+
+			Convey("then the expected content is returned", func() {
+				So(err, ShouldBeNil)
+				So(m.Count, ShouldEqual, 3)
+				So(m.Items, ShouldNotBeEmpty)
+				So(m.Items, ShouldHaveLength, 3)
+				mItem := m.Items[0]
+				So(mItem.URI, ShouldResemble, "/releases")
+				So(m.Limit, ShouldEqual, 3)
+				So(m.Offset, ShouldEqual, 0)
```
goofballLogic, line 636: This will pass even if offset is not specified in the Zebedee response (0 is the default value). Is that the intention?

zebedee/client_test.go
```diff
@@ -607,3 +608,95 @@ func TestClient_PublishedDataEndpoint(t *testing.T) {
...
+				So(m.TotalCount, ShouldEqual, 3)
+			})
+
+			Convey("and client.Do should be called once with the expected parameters", func() {
+				doCalls := httpClient.DoCalls()
+				So(doCalls, ShouldHaveLength, 1)
+				So(doCalls[0].Req.URL.Path, ShouldEqual, path)
+				//p := doCalls[0].Req.FormValue("offset")
```
goofballLogic, line 644: some comments left over here?

zebedee/client.go
```diff
@@ -469,6 +469,31 @@ func (c *Client) GetPublishedData(ctx context.Context, uriString string) ([]byte
...
+
+	if params != nil {
+		reqURL = fmt.Sprintf("%s?offset=%d",reqURL, params.Offset)
+		if params.Limit != nil {
+			reqURL = fmt.Sprintf("%s&limit=%d",reqURL, *(params.Limit))
```
goofballLogic, line 479: does it matter that this accepts arbitrarily large numbers (e.g. 9999999999 goes through)

zebedee/client_test.go
```diff
@@ -607,3 +608,95 @@ func TestClient_PublishedDataEndpoint(t *testing.T) {
...
+				doCalls := httpClient.DoCalls()
+				So(doCalls, ShouldHaveLength, 1)
+				So(doCalls[0].Req.URL.Path, ShouldEqual, path)
+				req := doCalls[0].Req
+				So(req.FormValue("offset"), ShouldEqual, "5")
+				So(req.FormValue("limit"), ShouldEqual, "9")
+			})
+		})
+	})
+
```
goofballLogic, line 678: Do 404 responses from Zebedee count as errors (in which case no extra test needed), or as empty collections?

zebedee/client.go
goofballLogic: I guess Limit could also be specified as 0 if that matters

zebedee/client_test.go
goofballLogic: Zebedee never returns 404 unless the wrong thing was requested

zebedee/client.go
goofballLogic: not checked at this level

zebedee/client_test.go
jessjenkins: TBH the paging variables are only there for future compatibility should we implement it. So no need to worry about it right now.

zebedee/client_test.go
jessjenkins: No 404 test is needed for the index as it's not specifying a specific resource to return.

rahulPalliyalil (approved):

PR #168, opened 2022-01-14
do not error on empty values when setting headers (for all headers)

DavidSubiros:
### What

Bugfix: 
When the headers error handling was fixed, it caused that errors setting headers were being returned to the caller.
Most callers assumed that empty values would be allowed, so that fix broke some expectations of callers, because by default, all headers would expect non-empty values, and fail with `ErrValueEmpty` otherwise.

In this PR we ignore `ErrValueEmpty` when trying to set any header, but the error type has been kept and returned by the lower level func `setRequestHeader` so that if a header becomes mandatory, we can just forward this error to the caller in the future.

### How to review

- Make sure unit tests pass

### Who can review

anyone

### DIFF
--- a/cantabular/client.go
+++ b/cantabular/client.go
@@ -136,8 +136,8 @@ func (c *Client) checkHealth(ctx context.Context, state *healthcheck.CheckState,
 		log.Error(ctx, "failed to request service health", err, logData)
 	} else {
 		code = res.StatusCode
-		defer closeResponseBody(ctx, res)
 	}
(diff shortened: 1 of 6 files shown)


redhug1 (changes requested):
The changes look good,
But, should the following also be fixed for cantabular:
cantabular/client.go:134:23: response body must be closed (bodyclose)
	res, err := c.httpGet(ctx, reqURL)
	                     ^

redhug1 (approved):

PR #169, opened 2022-01-17
Handle collection id when requesting bulletin data

rafahop:
### What

Handle collection id when requesting bulletin data

### How to review

Sense check. Tests pass

### Who can review

Anyone

### DIFF
--- a/zebedee/client.go
+++ b/zebedee/client.go
@@ -369,9 +369,7 @@ func (c *Client) GetCollection(ctx context.Context, userAccessToken, collectionI
 }
 
 // GetBulletin retrieves a bulletin from zebedee
-func (c *Client) GetBulletin(ctx context.Context, userAccessToken, lang, uri string) (Bulletin, error) {
-	collectionID := ""
-
+func (c *Client) GetBulletin(ctx context.Context, userAccessToken, collectionID, lang, uri string) (Bulletin, error) {
 	reqURL := c.createRequestURL(ctx, collectionID, lang, "/data", "uri="+uri)
 	b, _, err := c.get(ctx, userAccessToken, reqURL)
 	if err != nil {
--- a/zebedee/client_test.go
+++ b/zebedee/client_test.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


zebedee/client.go
```diff
@@ -369,9 +369,7 @@ func (c *Client) GetCollection(ctx context.Context, userAccessToken, collectionI
 }
 
 // GetBulletin retrieves a bulletin from zebedee
-func (c *Client) GetBulletin(ctx context.Context, userAccessToken, lang, uri string) (Bulletin, error) {
-	collectionID := ""
-
+func (c *Client) GetBulletin(ctx context.Context, userAccessToken, collectionID, lang, uri string) (Bulletin, error) {
```
bpathak-ons, line 372: just wondering what will happen if call is made with collectionID, should we include that scenario in the test too

zebedee/client_test.go
```diff
@@ -230,8 +230,10 @@ func TestUnitClient(t *testing.T) {
 	})
 
 	Convey("test GetBulletin", t, func() {
+		collectionId := ""
+		lang := "en"
 		Convey("returns the latest release of a bulletin", func() {
-			b, err := cli.GetBulletin(ctx, testAccessToken, "", "bulletin-latest-release")
+			b, err := cli.GetBulletin(ctx, testAccessToken, collectionId, lang, "bulletin-latest-release")
```
bpathak-ons, line 236: not sure about the requirement but should we consider the welsh lang too for a scenario testing

zebedee/client.go
rafahop: Thanks @bpathak-ons, that's a good point and I did consider passing a non empty string in the test but decided not to for simplicity. 
When we have a collection id, the zebedee URL [is prefixed with that value](https://github.com/ONSdigital/dp-api-clients-go/blob/9bc45000a35d64f1d14604377114892ba762b38c/zebedee/client.go#L440-L442). The current tests are mocking an http server whose [serving routes](https://github.com/ONSdigital/dp-api-clients-go/blob/9bc45000a35d64f1d14604377114892ba762b38c/zebedee/client_test.go#L34) could be easily amended to accept this new url but asserting the values wouldn't be straightforward. Especially since none of the other existing tests currently do anything around that and that change would mean a larger diff. 
I will be happy to look into it if we think there is value in it?

zebedee/client_test.go
rafahop: Similarly, there is nothing in the other existing tests to assert the value set as language when it reaches the mocked server but we could try and amend them if necessary?

bpathak-ons (dismissed):
looks good to me in view of the PR conversations

bpathak-ons (approved):

PR #170, opened 2022-01-19
Add client for articles API

rafahop:
### What

Add client for articles API

### How to review

Sense check. Test coverage.

### Who can review

Anyone

### DIFF
--- a/articles/articles.go
+++ b/articles/articles.go
@@ -1,0 +1,77 @@
+package articles
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	dperrors "github.com/ONSdigital/dp-api-clients-go/v2/errors"
+	"github.com/ONSdigital/dp-api-clients-go/v2/headers"
+	"github.com/ONSdigital/log.go/v2/log"
+)
+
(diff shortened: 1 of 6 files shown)


articles/articles.go
```diff
@@ -0,0 +1,77 @@
...
+	b, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, dperrors.New(
+			fmt.Errorf("failed to read response body from Articles API: %s", err),
+			resp.StatusCode,
+			nil,
+		)
+	}
+
+	if len(b) == 0 {
```
PatrickConnollyONS, line 54: Is this not a bit odd - it looks like we are goign to force a json.Unmarshal error? Why not just return an error taht describes what happened, like:
dperrors.New(
			"empty response"
			http.StatusInternalServerError,
			nil},
		)

rafahop: It does look odd indeed and I asked myself the same question. 
I just followed the pattern in this library as used in other sections but I think the idea of this is to explicitly output that it came empty rather than displaying an empty string in the error. The unmarshaling will fail anyway when `b` is empty.

We could remove it if we feel it is pointless. Otherwise, I think I should add a test case for it

PatrickConnollyONS (approved):

PR #171, opened 2022-01-20
Added localise code to request headers

joegasewicz:
### What

The areas profile needs to hit the areas profile api with an the language code via the headers.

### How to review

Check that the added AcceptedLang func & tests are valid.

### Who can review
404 Team


joegasewicz:
I think I should be passing through the current Accepted Headers (e.g en-GB,en-US;q=0.9,en;q=0.8) from the client -> dp-front-area-profiles -> dp-api-clients-go.

daiLlew (approved):
LGTM

PR #172, opened 2022-01-26
add is_cantabular_geograpy to codelist model

andre-urbani:
### What

Add is_cantabular_geography to CodeList model
Trello - https://trello.com/c/Y5cFREbZ/5360-changes-to-import-process-for-flexible-geographies-5

### How to review

Confirm changes are correct

### Who can review

Anyone

### DIFF
--- a/dataset/data.go
+++ b/dataset/data.go
@@ -82,10 +82,11 @@ type Event struct {
 
 // CodeList holds one of the codelists corresponding to a new Instance
 type CodeList struct {
-	ID          string `json:"id"`
-	HRef        string `json:"href"`
-	Name        string `json:"name"`
-	IsHierarchy bool   `json:"is_hierarchy"`
+	ID                    string `json:"id"`
+	HRef                  string `json:"href"`
+	Name                  string `json:"name"`
+	IsHierarchy           bool   `json:"is_hierarchy"`
+	IsCantabularGeography bool   `json:"is_cantabular_geography"`
 }
 
 // Version represents a version within a dataset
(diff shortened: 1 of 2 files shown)


dataset/data.go
```diff
@@ -82,10 +82,11 @@ type Event struct {
...
 type CodeList struct {
-	ID          string `json:"id"`
-	HRef        string `json:"href"`
-	Name        string `json:"name"`
-	IsHierarchy bool   `json:"is_hierarchy"`
+	ID                    string `json:"id"`
+	HRef                  string `json:"href"`
+	Name                  string `json:"name"`
+	IsHierarchy           bool   `json:"is_hierarchy"`
+	IsCantabularGeography bool   `json:"is_cantabular_geography"`
```
eldeal, line 89: I don't actually think this field (or isHierarchy) are valid in the dataset model for CodeList. 

Dataset API responses only link to the CodeList. This struct seems to be used in place of Dimension, in the []Dimension returned on a /version or /instance endpoint. e.g. https://api.beta.ons.gov.uk/v1/datasets/wellbeing-quarterly/editions/time-series/versions/3 

those responses look like this:
```
{
"href": "https://api.beta.ons.gov.uk/v1/code-lists/yyyy-qq",
"id": "yyyy-qq",
"label": "Time",
"links": {
"code_list": {},
"options": {},
"version": {}
},
"name": "time"
},
```

and don't currently contain. the isHierarchy field. If we need to be able to store this data in the dataset API, then I'd suggest removing this line from this PR and making the change to the Dataset API to allow the field first - then come back and add this when it's a recognized field :)

eldeal (changes requested):
Remove the bool from dataset/data.go, I don't think we should be adding it (or at least not yet)

eldeal (approved):

PR #173, opened 2022-01-26
created GetGeographyDimensions, modified contracts, to allow each cal…

DavidSubiros:
### What

- Implement cantabular `GetGeographyDimensions`, with the grahpql query described in the Trello card.
- Added Request/Response structs in contracts.go, and created QueryData, with an encode func, to generalise the encoding of queries.

trello card: https://trello.com/c/tSdkHCF2/5452-get-area-types-implement-endpoint-8

### How to review

- Make sure unit tests pass

(info only) tested by using a real client in the unit test `TestGetGeographyDimensionsHappy`:
```
cantabularClient := cantabular.NewClient(
			cantabular.Config{
				Host:       "http://localhost:8491",
				ExtApiHost: "http://localhost:8492",
			},
			dphttp.NewClient(),
			nil,
		)
```

### Who can review

anyone


cantabular/contract.go
```diff
@@ -39,13 +39,36 @@ type StaticDatasetQuery struct {
 	Dataset StaticDataset `json:"dataset" graphql:"dataset(name: $name)"`
 }
 
+// GetDimensionsByNameRequest holds the request variables required from the
+// caller for making a request to obtain dimensions (Cantabular variables) by name
+// POST [cantabular-ext]/graphql
+type GetDimensionsByNameRequest struct {
```
goofballLogic, line 45: I get that we have a convention of one contract type per client method, but this struct doesn't even express an opinion about the variable being a name and consequently is identical to `GetDimensionOptionsRequest`. Is that ok?

cantabular/gql/data.go
```diff
@@ -26,7 +26,7 @@ type Node struct {
 	Label      string      `json:"label"`
 	Categories Categories  `json:"categories"`
 	MapFrom    []Variables `json:"mapFrom"`
-	FilterOnly string      `json:"filterOnly"`
+	FilterOnly string      `json:"filterOnly,omitempty"`
```
goofballLogic, line 29: this is just for my own learning: why are some of these values in `gql` package and others are in `cantabular` package (`contract.go`)? Both end up being part of the client's "contract" don't they?

cantabular/static_dataset_test.go
```diff
@@ -422,6 +422,42 @@ func TestGetDimensionsByNameHappy(t *testing.T) {
...
+
+		cantabularClient := cantabular.NewClient(
+			cantabular.Config{
+				Host:       "cantabular.host",
+				ExtApiHost: "cantabular.ext.host",
+			},
+			mockHttpClient,
+			nil,
+		)
+
```
goofballLogic, line 446: Missing a test to verify that the query is correct

cantabular/static_dataset_test.go
```diff
@@ -773,6 +809,99 @@ var expectedDimensions = cantabular.GetDimensionsResponse{
...
+					]
+				},
+				"name": "Region"
+			}
+		}
+	}
+}
+`
+
+var expectedGeographyDimensions = cantabular.GetGeographyDimensionsResponse{
```
goofballLogic, line 861: as far as I can see, we're only round-tripping in the test - do we need all this detail?

cantabular/static_dataset_test.go
```diff
@@ -422,6 +422,42 @@ func TestGetDimensionsByNameHappy(t *testing.T) {
 	})
 }
```
goofballLogic, line 424: Missing UnhappyPath tests

goofballLogic (commented):
Few things (which may or may not be relevant)

cantabular/contract.go
DavidSubiros: It makes sense, I will replace 'variables' with 'names' in this struct (from the caller's point of view, it doesn't matter if it's encoded as 'variables' in the query, it is just a list of names for the dimensions that we want to retrieve).

cantabular/gql/data.go
DavidSubiros: I tried to separate the 'common' types from the contract, but maybe you are right and these types can be part of the contract.

cantabular/static_dataset_test.go
DavidSubiros: I used real responses for the tests, maybe I can move them to a separate file, so that the testing code is cleaner?

cantabular/gql/data.go
DavidSubiros: Maybe I will leave it as it is, as this can be used by multiple different generic graphql respones. But feel free to change it in the future.

cantabular/static_dataset_test.go
DavidSubiros: Yes, in fact, this was missing for all the tests. I added a few aux funcs to help create mocks and validate queries.

cantabular/static_dataset_test.go
DavidSubiros: done

cantabular/static_dataset_test.go
goofballLogic: dunno - if that's the way we generally document expected responses etc. that's fine. Could be cleaner in a new file maybe but meh...

PR #177, opened 2022-01-31
Refactored away from "population type" to "dataset"

goofballLogic:
### What

Following further discussion with the team, @ChrysmOre recommends that we refactor to "dataset" from "population type" with the insight that conventions within the existing Cantabular package are to stick to cantabular terms, allowing consumers to adopt ONS terminology (i.e. population type)

### How to review

Review simple refactoring. Check no "population type" nomenclature has been left behind

### Who can review

@ChrysmOre

### DIFF
--- a/cantabular/dataset_list.go
+++ b/cantabular/dataset_list.go
@@ -6,16 +6,16 @@ import (
 	"github.com/shurcooL/graphql"
 )
 
-type DatasetListItem struct {
+type ListDatasetsListItem struct {
 	Name graphql.String
 }
 
-type DatasetListQuery struct {
-	Datasets []DatasetListItem
+type ListDatasetsQuery struct {
+	Datasets []ListDatasetsListItem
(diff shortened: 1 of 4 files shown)


cantabular/dataset_list.go
```diff
@@ -6,16 +6,16 @@ import (
...
 )
 
-type PopulationTypeQueryDataset struct {
+type DatasetListItem struct {
 	Name graphql.String
 }
 
-type PopulationTypeQuery struct {
-	Datasets []PopulationTypeQueryDataset
+type DatasetListQuery struct {
```
ChrysmOre, line 13: Just one small thing.. Could we keep more precise naming of the related objects.. i.e. if the query is called ListDatasets name the query, items tests etc ListDatasetsxxx rather than DatasetListxxx

ChrysmOre (approved):

PR #179, opened 2022-02-01
add is_based_on field to dataset details struct

andre-urbani:
### What

Add is_based_on field to DatasetDetails struct
Trello - https://trello.com/c/CciD70lI/5473-populate-isbasedon-during-dataset-creation-5

### How to review

Confirm changes are correct

### Who can review

Anyone


DavidSubiros (approved):
LGTM
Validated that model is used both for `GET` and `PUT` endpoints, so it will be available for write and read.

PR #188, opened 2022-02-16
Create release model for release calendar api

rafahop:
### What

Create release model for the release calendar API

This is very similar to the zebedee model but stripping out non-required fields in the description

### How to review

Check model looks correct

### Who can review

Anyone


releasecalendar/data.go
```diff
@@ -0,0 +1,49 @@
+package releasecalendar
+
+// Release represents a release
+type Release struct {
+	URI                       string              `json:"uri"`
```
nshumoogum, line 5: Generally I would order the fields alphabetically, it is easier to find the field you are looking for that way.

The only reason not to do it alphabetically is if there is a logical order, e.g. grouping fields together. Here I think alphabetically works best

releasecalendar/data.go
```diff
@@ -0,0 +1,49 @@
+package releasecalendar
+
+// Release represents a release
+type Release struct {
```
nshumoogum, line 4: Should you include metadata fields like:

```go
LastUpdated       time.Time                         `json:"last_updated,omitempty"`
```

`omitempty` if the field doesn't exist, but once releases are ported over into document db, I would expect this field to be set. Maybe this is future work

nshumoogum: Another field would be ETag - https://github.com/ONSdigital/dp-dataset-api/blob/develop/models/instance.go#L25

nshumoogum (approved):

PR #190, opened 2022-02-22
Feature/add dataset struct to model

mr-nick17:
### What

Added `dataset` struct to model so dataset details are returned on the filter model

### How to review

Sense check

### Who can review

Anyone but me


rafahop (approved):
LGTM

PR #196, opened 2022-02-25
Add population-type to CreateBlueprint filter model

franmoore05:
### What

Add population-type to CreateBlueprint filter model for Cantabular filters.

### How to review

Ensure the changes make sense.

### Who can review

Anyone.

### DIFF
--- a/filter/data.go
+++ b/filter/data.go
@@ -41,9 +41,10 @@ type DimensionOptions struct {
 
 // CreateBlueprint represents the fields required to create a filter blueprint
 type CreateBlueprint struct {
-	Dataset    Dataset          `json:"dataset"`
-	Dimensions []ModelDimension `json:"dimensions"`
-	FilterID   string           `json:"filter_id"`
+	Dataset        Dataset          `json:"dataset"`
+	Dimensions     []ModelDimension `json:"dimensions"`
+	FilterID       string           `json:"filter_id"`
+	PopulationType string           `json:"population_type"`
 }
 
 // Dataset represents the dataset fields required to create a filter blueprint


filter/data.go
```diff
@@ -41,9 +41,10 @@ type DimensionOptions struct {
 
 // CreateBlueprint represents the fields required to create a filter blueprint
 type CreateBlueprint struct {
-	Dataset    Dataset          `json:"dataset"`
-	Dimensions []ModelDimension `json:"dimensions"`
-	FilterID   string           `json:"filter_id"`
+	Dataset        Dataset          `json:"dataset"`
+	Dimensions     []ModelDimension `json:"dimensions"`
+	FilterID       string           `json:"filter_id"`
+	PopulationType string           `json:"population_type"`
```
mr-nick17, line 47: Should `PopulationType` has omitempty? This method is used my CMD
```suggestion
	PopulationType string          `json:"population_type,omitempty"`
```

franmoore05: Indeed it should, I've now added it.

ChrysmOre:
We're going to also need to update the CreateBlueprint function itself to use the changes to the request.. By the looks of things it's actually not fit for purpose as it doesn't add dimension-options to each dimension.. Looks like this will be a bigger change than anticipated.. We'll need to break the function signature to include population_type and dimension options, and add logic to add them if present.. It would be so much better if these API Client functions used structs as their inputs rather than 20 arguments to a function.. These changes would be a lot less breaking

franmoore05:
This has been superseded by another PR

PR #201, opened 2022-03-03
Feature/adding code for get areas endpoint

franmoore05:
### What

Added functions and models to allow graphql queries to be made against Cantabular for an areas search.

### How to review

Ensure the changes make sense.

### Who can review

Anyone.

### DIFF
--- a/cantabular/dimensions.go
+++ b/cantabular/dimensions.go
@@ -2,7 +2,8 @@ package cantabular
 
 import (
 	"context"
-	"errors"
+
+	"github.com/pkg/errors"
 
 	"github.com/ONSdigital/dp-api-clients-go/v2/cantabular/gql"
 	dperrors "github.com/ONSdigital/dp-api-clients-go/v2/errors"
@@ -155,13 +156,8 @@ func (c *Client) GetAreas(ctx context.Context, req QueryData) (*GetAreasResponse
 		Errors []gql.Error      `json:"errors,omitempty"`
 	}{}
 
-	data := QueryData{
-		Dataset: req.Dataset,
-		Text:    req.Text,
-	}
-
-	if err := c.queryUnmarshal(ctx, QueryAreasByArea, data, resp); err != nil {
-		return nil, err
+	if err := c.queryUnmarshal(ctx, QueryAreasByArea, req, resp); err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshal query")
 	}
 
 	if resp != nil && len(resp.Errors) != 0 {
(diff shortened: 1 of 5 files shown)


cantabular/dimensions.go
```diff
@@ -148,3 +148,29 @@ func (c *Client) GetDimensionOptions(ctx context.Context, req GetDimensionOption
...
+	resp := &struct {
+		Data   GetAreasResponse `json:"data"`
+		Errors []gql.Error      `json:"errors,omitempty"`
+	}{}
+
+	data := QueryData{
+		Dataset: req.Dataset,
+		Text:    req.Text,
+	}
+
```
PauloPortugal, line 162: Why do we need to instantiate `data` since you already have the data on `req`?

You could pass it to the private function, since the `GetAreas(ctx context.Context, req QueryData)` is already documenting that it is of type `QueryData`
```
if err := c.queryUnmarshal(ctx, QueryAreasByArea, req, resp); err !=
```

cantabular/dimensions.go
```diff
@@ -148,3 +148,29 @@ func (c *Client) GetDimensionOptions(ctx context.Context, req GetDimensionOption
...
+	if err := c.queryUnmarshal(ctx, QueryAreasByArea, data, resp); err != nil {
+		return nil, err
```
PauloPortugal, line 164: speaking with Josh on my ticket, he was keen to always wrap the error with a contextual message. Not sure if you want to do that.
Non-blocker.

PauloPortugal (dismissed):
Looks good to me from a technical perspective.
Just a small remark on the `/GET` endpoint, but that is a non-blocker.

cantabular/dimensions.go
ChrysmOre: Yes, please when returning errors can we always make sure to wrap with a brief description of what has failed.. Historically we've done this with either `fmt.Errorf` or `errors.Wrap` (`github.com/pkg/errors`), now moving to only using `errors.Wrap` in order to retain a stack trace. Example:
```
return nil, errors.Wrap(err, "failed to unmarshal query")
```

cantabular/dimensions.go
ChrysmOre: I think that's sensible ^

ChrysmOre (changes requested):
Just the error wrapping

cantabular/dimensions.go
franmoore05: This has now been updated

cantabular/dimensions.go
franmoore05: I've updated this and also updated the tests to reflect the new way of wrapping errors

PauloPortugal (approved):

ChrysmOre (approved):
LGTM

PR #204, opened 2022-03-08
add CreateFilter function and unit tests with population type

mohsdev:
### What

### How to review

### Who can review

### DIFF
--- a/filter/data.go
+++ b/filter/data.go
@@ -21,7 +21,7 @@ type Dimensions struct {
 // Dimension represents a dimension response from the filter api
 type Dimension struct {
 	Name string `json:"name"`
-	URI  string `json:"dimension_url"`
+	URI  string `json:"dimension_url,omitempty"`
 }
 
 // DimensionOption represents a dimension option from the filter api
@@ -44,6 +44,18 @@ type CreateBlueprint struct {
 	Dataset    Dataset          `json:"dataset"`
 	Dimensions []ModelDimension `json:"dimensions"`
 	FilterID   string           `json:"filter_id"`
+}
+
+// CreateFilterRequest represents the fields required to create a filter
+type CreateFilterRequest struct {
+	Dataset        Dataset     `json:"dataset"`
+	Dimensions     []Dimension `json:"dimensions"`
+	PopulationType string      `json:"population_type"`
+}
+
+// CreateFilterResponse is the response body for POST /filters
+type CreateFilterResponse struct {
+	FilterID string `json:"filter_id"`
 }
 
 // Dataset represents the dataset fields required to create a filter blueprint
--- a/filter/data.go
+++ b/filter/data.go
@@ -21,7 +21,7 @@ type Dimensions struct {
 // Dimension represents a dimension response from the filter api
 type Dimension struct {
 	Name string `json:"name"`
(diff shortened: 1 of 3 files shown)


filter/filter.go
```diff
@@ -1084,3 +1084,81 @@ func (c *Client) doPatchWithAuthHeaders(ctx context.Context, userAuthToken, serv
...
+	data, err := json.Marshal(filter)
+	if err != nil {
+		return "", "", err
+	}
+	uri := c.hcCli.URL + "/filters"
+	clientlog.Do(ctx, "attempting to create filter", service, uri, log.Data{
+		"method":    "POST",
+		"datasetID": datasetID,
+		"edition":   edition,
+		"version":   version,
```
franmoore05, line 1119: We should also log out the population-type here

mohsdev: done

franmoore05 (approved):
LGTM

filter/data.go
```diff
@@ -21,7 +21,7 @@ type Dimensions struct {
 // Dimension represents a dimension response from the filter api
 type Dimension struct {
 	Name string `json:"name"`
-	URI  string `json:"dimension_url"`
+	URI  string `json:"dimension_url,omitempty"`
```
PauloPortugal, line 24: Why did we change it?
Are we absolutely sure this won't be required when serializing?

filter/filter.go
```diff
@@ -1084,3 +1084,82 @@ func (c *Client) doPatchWithAuthHeaders(ctx context.Context, userAuthToken, serv
 	// do the request
 	return c.hcCli.Client.Do(ctx, req)
 }
+
+// CreateFilter creates a filter and returns the associated filterID and eTag
+func (c *Client) CreateFilter(ctx context.Context, userAuthToken, serviceAuthToken, downloadServiceToken, collectionID, datasetID, edition, version, populationType string, names []string) (filterID, eTag string, err error) {
```
PauloPortugal, line 1089: It would have been better to create a model that represents all the required data to create a filter.
This would avoid us having to refactor method signatures in the future. 

I have also realise that this is what we are doing for other methods.

Just a comment.

PauloPortugal (changes requested):
It looks good, although I don't like the long methods and the style, but it is within what is happening on the same package.

A couple of things:
 * The next time please add more info on the PR, i.e, replace the PR template with useful info for the developers reviewing.
 * Your first commit was pushed via `root`. As a rule of thumb avoid running commands as root on your machine. You won't be able to merge this PR until you amend that commit or fix it locally, because that commit is not signed. Please change this.

filter/data.go
mohsdev: sorry, let me make it unchanged

mohsdev: @PauloPortugal i already commited a fix just now with "-S", help to check if the GPG key sign works

PauloPortugal: Your first commit is still showing as being created by `root`. you need to fix that.

goofballLogic (changes requested):
Various things wrong with this as discussed elsewhere

PauloPortugal (dismissed):

goofballLogic:
Superceded by #207

PR #220, opened 2022-03-21
Add PopulationType field to Filter Model

jayfreestone:
### What

Adds the `PopulationType` field [provided by](https://github.com/ONSdigital/dp-cantabular-filter-flex-api/blob/ed2eb6652e835d0532440b0e47acb3e02305c0b3/swagger.yaml#L531) the Cantabular Filter-Flex API. I'm presuming adding a field here and using `omitempty` should be fine/compatible with CMD.

We need this in order to determine the underlying Cantabular dataset to pass to `GetGeographyDimensions` for [5569](https://trello.com/c/yOjZDYAj/5569-create-front-end-to-display-list-of-area-types). While it's possible to determine with an additional call to the Dataset API, since the data is already here we may as well avoid an extra request.

### How to review

Double check the new field matches the API spec.

### Who can review

Anyone.

### DIFF
--- a/filter/data.go
+++ b/filter/data.go
@@ -74,18 +74,19 @@ type Dataset struct {
 
 // Model represents a model returned from the filter api
 type Model struct {
-	FilterID    string              `json:"filter_id"`
-	InstanceID  string              `json:"instance_id"`
-	Links       Links               `json:"links"`
-	DatasetID   string              `json:"dataset_id"`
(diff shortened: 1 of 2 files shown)


franmoore05 (approved):
LGTM

jayfreestone:
Closing as the new field has been added [here](https://github.com/ONSdigital/dp-api-clients-go/commit/bbef03ff3742f893f24358ffe3ed70148ebc17bf).

PR #229, opened 2022-03-25
Removes getAncestor

joegasewicz:
### What

Remove Ancestor struct & GetAncestors

We don't have a requirement for a GetAncestors func anymore as this data is returned in the `/areas` response under the `ancestors` key.

We also don't require the Ancestors struct as we can use `[]Ancestor` instead without the double nested `Ancestor.Ancestor` struct member access notation. 

### How to review
Run the tests

Run: `make test`

### Who can review
anyone
myself


pidah (approved):
LGTM

PR #232, opened 2022-03-29
Add 'SubmitFilter' function

PauloPortugal:
### What

A new endpoint is required in dp-cantabular-filter-flex-api to submit
the user choices to generate the cantabular downloads, so we need to
create an associated function in dp-api-clients-go.

A new function should be created to handle the request to submit a
filter for a cantabular dataset which should POST to
`/filters/{filterid}/submit` in `dp-cantabular-filter-flex-api`.

Resolves: [5614](https://trello.com/c/JnvjQOa6/5614-create-function-in-dp-api-clients-go-for-filter-submit)

### How to review

* Check the Trello story 
* Inspect the code
* Run the unit tests

### Who can review

Any **Team B - Get Data** developer

### DIFF
--- a/cantabular/filter.go
+++ b/cantabular/filter.go
@@ -1,0 +1,128 @@
+package cantabular
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/ONSdigital/dp-api-clients-go/v2/clientlog"
+	"github.com/ONSdigital/dp-api-clients-go/v2/filter"
+	"github.com/ONSdigital/dp-api-clients-go/v2/headers"
+	"github.com/ONSdigital/log.go/v2/log"
+)
+
+const (
+	service     = "dp-cantabular-filter-flex-api"
+	POST_METHOD = "POST"
+)
+
+type SubmitFilterRequest struct {
+	FilterID       string                    `json:"filter_id"`
+	Dimensions     []filter.DimensionOptions `json:"dimension_options,omitempty"`
+	PopulationType string                    `json:"population_type"`
+}
+
+type Event struct {
+	Timestamp time.Time `bson:"timestamp" json:"timestamp"`
+	Name      string    `bson:"name"      json:"name"`
+}
+
+type SFDataset struct {
+	ID      string `bson:"id"      json:"id"`
+	Edition string `bson:"edition" json:"edition"`
+	Version int    `bson:"version" json:"version"`
+}
+
+type Link struct {
+	HREF string `bson:"href"           json:"href"`
+	ID   string `bson:"id,omitempty"   json:"id,omitempty"`
+}
+
+type Links struct {
+	Version Link `bson:"version" json:"version"`
+	Self    Link `bson:"self"    json:"self"`
+}
+
+type SFDimension struct {
+	Name         string   `bson:"name"          json:"name"`
+	Options      []string `bson:"options"       json:"options"`
+	DimensionURL string   `bson:"dimension_url" json:"dimension_url,omitempty"`
+	IsAreaType   bool     `bson:"is_area_type"  json:"is_area_type"`
+}
+
+type SubmitFilterResponse struct {
+	InstanceID       string        `json:"instance_id"`
+	DimensionListUrl string        `json:"dimension_list_url"`
+	FilterID         string        `json:"filter_id"`
+	Events           []Event       `json:"events"`
+	Dataset          SFDataset     `json:"dataset"`
+	Links            Links         `json:"links"`
+	PopulationType   string        `json:"population_type"`
+	Dimensions       []SFDimension `json:"dimensions"`
+}
+
+// SubmitFilter function to submit the request to submit a filter for a cantabular dataset.
+// Should POST to /filters/{filterid}/submit in dp-cantabular-filter-flex-api microservice.
+func (c *Client) SubmitFilter(ctx context.Context, userAuthToken, serviceAuthToken, downloadServiceToken, ifMatch string, sfr SubmitFilterRequest) (*SubmitFilterResponse, string, error) {
+	b, err := json.Marshal(sfr)
+	if err != nil {
+		return nil, "", err
+	}
+
+	uri := fmt.Sprintf("%s/filters/%s/submit", c.extApiHost, sfr.FilterID)
+
+	clientlog.Do(ctx, "updating filter job", service, uri, log.Data{
+		"method": POST_METHOD,
+		"body":   string(b),
+	})
+
+	req, err := http.NewRequest(POST_METHOD, uri, bytes.NewBuffer(b))
+	if err != nil {
+		return nil, "", err
+	}
+
+	if err = headers.SetAuthToken(req, userAuthToken); err != nil {
+		return nil, "", fmt.Errorf("failed to set auth token: %w", err)
+	}
+	if err = headers.SetServiceAuthToken(req, serviceAuthToken); err != nil {
+		return nil, "", fmt.Errorf("failed to set service auth token: %w", err)
+	}
+	if err = headers.SetDownloadServiceToken(req, downloadServiceToken); err != nil {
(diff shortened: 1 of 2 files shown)


ChrysmOre (changes requested):
Ah,, You've put the code in the wrong API Client. The Cantabular client is for directly querying Cantabular. Functions that need to call filter-api or cantabular-filter-flex-api need to be in `filter` of `filterflex` api clients respectively

PauloPortugal:
@ChrysmOre this is not what was suggested by @franmoore05 on the [ticket](https://trello.com/c/JnvjQOa6/5614-create-function-in-dp-api-clients-go-for-filter-submit).

Apologies, this is probably a case where a three-amigos would have picked this one up a lot earlier.
I am with Andrew finishing a ticket and will look into this after that ASAP.

PauloPortugal:
@ChrysmOre  & @franmoore05 please review this again.

filter/filter.go
```diff
@@ -606,6 +610,67 @@ func (c *Client) UpdateBlueprint(ctx context.Context, userAuthToken, serviceAuth
 	return m, eTag, nil
 }
 
+// SubmitFilter function to submit the request to submit a filter for a cantabular dataset.
+// Should POST to /filters/{filterid}/submit in dp-cantabular-filter-flex-api microservice.
+func (c *Client) SubmitFilter(ctx context.Context, userAuthToken, serviceAuthToken, downloadServiceToken, ifMatch string, sfr SubmitFilterRequest) (*SubmitFilterResponse, string, error) {
+	b, err := json.Marshal(sfr)
+	if err != nil {
+		return nil, "", err
```
ChrysmOre, line 618: Please remember to wrap errors
```suggestion
		return nil, "", errors.Wrap(err, "failed to marshal request)
```

filter/filter.go
```diff
@@ -606,6 +610,67 @@ func (c *Client) UpdateBlueprint(ctx context.Context, userAuthToken, serviceAuth
...
+	uri := fmt.Sprintf("%s/filters/%s/submit", c.hcCli.URL, sfr.FilterID)
+
+	clientlog.Do(ctx, "updating filter job", service, uri, log.Data{
+		"method": postMethod,
+		"body":   string(b),
+	})
+
+	req, err := http.NewRequest(postMethod, uri, bytes.NewBuffer(b))
+	if err != nil {
+		return nil, "", err
```
ChrysmOre, line 630: (and here)

filter/filter.go
```diff
@@ -606,6 +610,67 @@ func (c *Client) UpdateBlueprint(ctx context.Context, userAuthToken, serviceAuth
...
+	}
+
+	if err = headers.SetAuthToken(req, userAuthToken); err != nil {
+		return nil, "", fmt.Errorf("failed to set auth token: %w", err)
```
ChrysmOre, line 634: We should be using (github.com/pkg/errors) .Wrap() rather than fmt.Errorf to preserve the stack trace. 
```suggestion
		return nil, "", errors.wrap(err, "failed to set auth token")
```

filter/filter.go
```diff
@@ -606,6 +610,67 @@ func (c *Client) UpdateBlueprint(ctx context.Context, userAuthToken, serviceAuth
...
+	resp, err := c.hcCli.Client.Do(ctx, req)
+	if err != nil {
+		return nil, "", err
+	}
+	defer closeResponseBody(ctx, resp)
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, "", filter.ErrInvalidFilterAPIResponse{ExpectedCode: http.StatusOK, ActualCode: resp.StatusCode, URI: uri}
+	}
+
```
ChrysmOre, line 655: Several more places errors aren't wrapped

filter/data.go
```diff
@@ -133,3 +137,20 @@ type Preview struct {
...
 	Rows            [][]string `json:"rows"`
 }
+
+type SubmitFilterRequest struct {
+	FilterID       string                    `json:"filter_id"`
+	Dimensions     []filter.DimensionOptions `json:"dimension_options,omitempty"`
+	PopulationType string                    `json:"population_type"`
+}
+
+type SubmitFilterResponse struct {
```
ChrysmOre, line 147: You should be aware the main filter response (and I imagine submitFilter will follow suit) response model is about to change to be inline with actual filter-api behaviour (it turns out the swagger spec was very out of date) so you might have to come back and update this.

filter/filter.go
PauloPortugal: This has been addressed.

filter/filter.go
PauloPortugal: This has been addressed.

filter/filter.go
PauloPortugal: This has been addressed.

filter/filter.go
PauloPortugal: This has been addressed.

filter/data.go
PauloPortugal: @ChrysmOre dully noted. 
@franmoore05 should this be addressed as part of this ticket?

filter/filter.go
```diff
@@ -4,22 +4,26 @@ import (
...
 	health "github.com/ONSdigital/dp-healthcheck/healthcheck"
 	dprequest "github.com/ONSdigital/dp-net/request"
 	"github.com/ONSdigital/log.go/v2/log"
+	"github.com/pkg/errors"
 )
 
-const service = "filter-api"
+const (
+	service    = "filter-api"
+	postMethod = "POST"
```
ChrysmOre, line 25: This const is already part of the http package `http.MethodPost`

filter/filter.go
```diff
@@ -606,6 +610,71 @@ func (c *Client) UpdateBlueprint(ctx context.Context, userAuthToken, serviceAuth
...
+	if err = headers.SetDownloadServiceToken(req, downloadServiceToken); err != nil {
+		return nil, "", errors.Wrap(err, "failed to set download service token")
+	}
+	if err = headers.SetIfMatch(req, ifMatch); err != nil {
+		return nil, "", errors.Wrap(err, "failed to set if match")
+	}
+
+	resp, err := c.hcCli.Client.Do(ctx, req)
+	if err != nil {
+		return nil, "", errors.Wrap(err, "failed to create a new http client")
```
ChrysmOre, line 648: I don't think that is creating a http client, but making the request?

filter/filter.go
```diff
@@ -606,6 +610,71 @@ func (c *Client) UpdateBlueprint(ctx context.Context, userAuthToken, serviceAuth
...
+	}
+	defer closeResponseBody(ctx, resp)
+
+	if resp.StatusCode != http.StatusOK {
```
ChrysmOre, line 652: Would it be better to move this check till after we have the response body, and then return the response in the error logData?

filter/filter.go
PauloPortugal: this has been addressed.

filter/filter.go
```diff
@@ -606,6 +610,71 @@ func (c *Client) UpdateBlueprint(ctx context.Context, userAuthToken, serviceAuth
 	return m, eTag, nil
 }
 
+// SubmitFilter function to submit the request to submit a filter for a cantabular dataset.
+// Should POST to /filters/{filterid}/submit in dp-cantabular-filter-flex-api microservice.
+func (c *Client) SubmitFilter(ctx context.Context, userAuthToken, serviceAuthToken, downloadServiceToken, ifMatch string, sfr SubmitFilterRequest) (*SubmitFilterResponse, string, error) {
+	b, err := json.Marshal(sfr)
+	if err != nil {
+		return nil, "", errors.Wrap(err, "could not unmarshal")
```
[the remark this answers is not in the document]
PauloPortugal: Terrible Copy/paste skills.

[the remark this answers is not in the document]
PauloPortugal: This has been addressed.

filter/filter.go
PauloPortugal: This has been addressed.

ChrysmOre (approved):
LGTM

PR #234, opened 2022-04-01
Handle slightly odd behaviour with zip file readers for upload client

mryan321:
File readers for zip files behave a little differently than normal readers: https://github.com/golang/go/issues/32858


braddle (approved):
LGTM

PR #235, opened 2022-04-01
Add updateDimensions to filters client

jackrehaag:
### What

Adds `updateDimensions` to send `PUT` requests to the filters api.

### How to review

Read the code and tests.

### Who can review

Anyone.

### DIFF
--- a/filter/filter.go
+++ b/filter/filter.go
@@ -853,6 +853,67 @@ func (c *Client) PatchDimensionValues(ctx context.Context, userAuthToken, servic
 	return latestETag, nil
 }
 
+func (c *Client) UpdateDimensions(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, id, name, ifMatch string, dimension Dimension) (dim Dimension, eTag string, err error) {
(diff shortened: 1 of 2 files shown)


PauloPortugal (approved):
LGTM

filter/filter.go
```diff
@@ -853,6 +853,67 @@ func (c *Client) PatchDimensionValues(ctx context.Context, userAuthToken, servic
...
+	respBody, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return dimension, "", err
+	}
+
+	var updatedDimension Dimension
+
+	if err = json.Unmarshal(respBody, &updatedDimension); err != nil {
+		return updatedDimension, "", err
+	}
```
jayfreestone, line 912: Equivalent for our needs but for reference (if you don't need to re-use the body), you can use a JSON decoder to read from a stream:

```go
var updatedDimension Dimension
if err = json.NewDecoder(resp.Body).Decode(&updatedDimension); err != nil {
	return updatedDimension, "", err
}
```

PR #238, opened 2022-04-08
Add GetAreaTypeAreas method to Cantabular client (#5651)

jayfreestone:
### What

Adds a new method which allows for the search/retrieval of areas for a specific area type (e.g. all the areas for `City`).

Resolves [#5651](https://trello.com/c/PDCxskCW)

### How to review

Confirm the method follows the format/structure of the existing API methods and unit tests.

### Who can review

Anyone.

### DIFF
--- a/cantabular/dimensions_test.go
+++ b/cantabular/dimensions_test.go
@@ -459,15 +459,20 @@ func TestGetDimensionOptionsUnhappy(t *testing.T) {
 	})
 }
 
-func TestGetAreasHappy(t *testing.T) {
-	Convey("Given a correct getAreas response from the /graphql endpoint", t, func() {
+func TestGetAreas(t *testing.T) {
+	Convey("Given a valid response from the /graphql endpoint", t, func() {
+		const dataset = "Example"
+		const variable = "city"
+		const category = "london"
+
 		testCtx := context.Background()
 		mockHttpClient, cantabularClient := newMockedClient(mockRespBodyGetAreas, http.StatusOK)
 
 		Convey("When GetAreas is called", func() {
-			req := cantabular.QueryData{
-				Dataset: "Example",
-				Text:    "London",
+			req := cantabular.GetAreasRequest{
+				Dataset:  dataset,
+				Variable: variable,
+				Category: category,
 			}
 
 			resp, err := cantabularClient.GetAreas(testCtx, req)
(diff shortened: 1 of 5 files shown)


jayfreestone:
> Don't we already have a GetAreas function? I think it might of been better to update the existing one if all that was needed was to allow to specify specific areas to return

@ChrysmOre Yep, good point, this does effectively supersede `GetAreas`. I wasn't sure if I wanted to touch the existing method, since it's a breaking change and would require a full client version bump (different params, different return type since they're not flattened).

Happy to rename this one and replace the old one if we're confident it's not being used anywhere other than the Dimensions API, where it's being removed (https://github.com/ONSdigital/dp-cantabular-dimension-api/pull/12)

ChrysmOre:
> > Don't we already have a GetAreas function? I think it might of been better to update the existing one if all that was needed was to allow to specify specific areas to return
> 
> @ChrysmOre Yep, good point, this does effectively supersede `GetAreas`. I wasn't sure if I wanted to touch the existing method, since it's a breaking change and would require a full client version bump (different params, different return type since they're not flattened).
> 
> Happy to rename this one and replace the old one if we're confident it's not being used anywhere other than the Dimensions API, where it's being removed ([ONSdigital/dp-cantabular-dimension-api#12](https://github.com/ONSdigital/dp-cantabular-dimension-api/pull/12))

Yeah I'm fairly sure that's the only place it's being used, I'd say definitely update the existing method rather than create a new one

cantabular/dimensions_test.go
```diff
@@ -459,15 +459,20 @@ func TestGetDimensionOptionsUnhappy(t *testing.T) {
...
-	Convey("Given a correct getAreas response from the /graphql endpoint", t, func() {
+func TestGetAreas(t *testing.T) {
+	Convey("Given a valid response from the /graphql endpoint", t, func() {
+		const dataset = "Example"
+		const variable = "city"
+		const category = "london"
+
 		testCtx := context.Background()
-		mockHttpClient, cantabularClient := newMockedClient(mockRespBodyGetAreas, http.StatusOK)
+		mockHttpClient, cantabularClient := newMockedClient(mockRespBodyGetAreaTypeAreas, http.StatusOK)
```
jackrehaag, line 469: Tests still refer to `GetAreaTypesAreas` rather than `GetAreas`.

jackrehaag (approved):

PR #245, opened 2022-04-25
Add GetAreas to dimension client

jackrehaag:
### What

Adds `GetAreas` to the dimension client. Also deduplicated some code around creating the dimension requests.
Context:
[https://trello.com/c/HNH1I1Dj/5650-add-getareas-method-to-cantabular-dimensions-api-client](https://trello.com/c/HNH1I1Dj/5650-add-getareas-method-to-cantabular-dimensions-api-client)

### How to review

Check code and tests against ticket/

### Who can review

Anyone.

### DIFF
--- a/dimension/areas.go
+++ b/dimension/areas.go
@@ -1,0 +1,13 @@
+package dimension
+
+// Area is an area model with ID and Label
+type Area struct {
+	ID       string `json:"id"`
+	Label    string `json:"label"`
+	AreaType string `json:"area-type"`
+}
+
+// GetAreasResponse is the response object for GET /areas
+type GetAreasResponse struct {
+	Areas []Area `json:"areas"`
+}
--- a/dimension/client.go
+++ b/dimension/client.go
@@ -27,6 +27,14 @@ type Client struct {
 	baseURL *url.URL
 }
 
+type GetAreasInput struct {
+	UserAuthToken    string
+	ServiceAuthToken string
+	DatasetID        string
+	AreaTypeID       string
+	Text             string
+}
+
 // NewClient creates a new instance of Client with a given Dimensions API URL
 func NewClient(dimensionsAPIURL string) (*Client, error) {
 	client := health.NewClient(service, dimensionsAPIURL)
@@ -39,7 +47,7 @@ func NewWithHealthClient(hcCli *health.Client) (*Client, error) {
 	client := health.NewClientWithClienter(service, hcCli.URL, hcCli.Client)
 	baseURL, err := url.Parse(client.URL)
 	if err != nil {
-		return nil, fmt.Errorf("error parsing URL: %w", err)
+		return nil, errors.Wrap(err, "error parsing URL")
 	}
 
 	// The Parse method on `url.URL` uses a trailing slash to determine
@@ -56,13 +64,13 @@ func (c *Client) Checker(ctx context.Context, check *healthcheck.CheckState) err
 	return c.hcCli.Checker(ctx, check)
 }
 
-func (c *Client) createDimensionGetRequest(ctx context.Context, userAuthToken, serviceAuthToken, urlPath string, logData log.Data, urlValues url.Values) (*http.Request, error) {
+func (c *Client) createGetRequest(ctx context.Context, userAuthToken, serviceAuthToken, urlPath string, urlValues url.Values) (*http.Request, error) {
 	areasURL, err := c.baseURL.Parse(urlPath)
 	if err != nil {
 		return &http.Request{}, dperrors.New(
 			errors.Wrap(err, "failed to parse areas URL"),
 			http.StatusInternalServerError,
-			logData,
+			log.Data{},
 		)
 	}
 
@@ -74,20 +82,20 @@ func (c *Client) createDimensionGetRequest(ctx context.Context, userAuthToken, s
 		return &http.Request{}, dperrors.New(
 			errors.Wrap(err, "failed to create request"),
 			http.StatusBadRequest,
-			logData,
+			log.Data{},
 		)
 	}
 	return req, nil
 }
 
-func checkDimensionGetResponse(logData log.Data, resp http.Response) error {
+func checkGetResponse(resp *http.Response) error {
 	if resp.StatusCode == http.StatusNotFound {
 		var errorResp ErrorResp
 		if err := json.NewDecoder(resp.Body).Decode(&errorResp); err == nil {
 			return dperrors.New(
 				fmt.Errorf("error response from Dimensions API (%d): %w", resp.StatusCode, errorResp),
 				http.StatusInternalServerError,
-				logData,
+				log.Data{},
 			)
 		}
 	}
@@ -96,9 +104,9 @@ func checkDimensionGetResponse(logData log.Data, resp http.Response) error {
 		// Best effort — an empty body is fine for the error message
 		body, _ := io.ReadAll(resp.Body)
 		return dperrors.New(
-			fmt.Errorf("error response from Dimensions API (%d): %s", resp.StatusCode, body),
+			errors.Errorf("error response from Dimensions API (%d): %s", resp.StatusCode, body),
 			http.StatusInternalServerError,
-			logData,
+			log.Data{},
 		)
 	}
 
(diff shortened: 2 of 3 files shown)


dimension/client.go
```diff
@@ -55,36 +56,76 @@ func (c *Client) Checker(ctx context.Context, check *healthcheck.CheckState) err
...
+	}
+	return req, nil
+}
+
+func checkDimensionGetResponse(logData log.Data, resp http.Response) error {
+	if resp.StatusCode == http.StatusNotFound {
+		var errorResp ErrorResp
+		if err := json.NewDecoder(resp.Body).Decode(&errorResp); err == nil {
+			return dperrors.New(
+				fmt.Errorf("error response from Dimensions API (%d): %w", resp.StatusCode, errorResp),
```
jayfreestone, line 88: `errors.Wrapf`

dimension/client.go
```diff
@@ -55,36 +56,76 @@ func (c *Client) Checker(ctx context.Context, check *healthcheck.CheckState) err
...
+				logData,
+			)
+		}
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		// Best effort — an empty body is fine for the error message
+		body, _ := io.ReadAll(resp.Body)
+		return dperrors.New(
+			fmt.Errorf("error response from Dimensions API (%d): %s", resp.StatusCode, body),
```
jayfreestone, line 99: `errors.Wrapf`

dimension/client.go
```diff
@@ -100,37 +141,76 @@ func (c *Client) GetAreaTypes(ctx context.Context, userAuthToken, serviceAuthTok
...
 	}
 
-	if resp.StatusCode != http.StatusOK {
-		// Best effort — an empty body is fine for the error message
-		body, _ := io.ReadAll(resp.Body)
+	var areaTypes GetAreaTypesResponse
+	if err := json.NewDecoder(resp.Body).Decode(&areaTypes); err != nil {
 		return GetAreaTypesResponse{}, dperrors.New(
-			fmt.Errorf("error response from Dimensions API (%d): %s", resp.StatusCode, body),
+			fmt.Errorf("unable to deserialize area types response: %w", err),
```
jayfreestone, line 151: `errors.Wrapf`

dimension/client.go
```diff
@@ -100,37 +141,76 @@ func (c *Client) GetAreaTypes(ctx context.Context, userAuthToken, serviceAuthTok
...
+			logData,
+		)
+	}
+
+	clientlog.Do(ctx, "getting areas", service, req.URL.String(), logData)
+
+	resp, err := c.hcCli.Client.Do(ctx, req)
+	if err != nil {
+		return GetAreasResponse{}, dperrors.New(
+			fmt.Errorf("failed to get response from Dimensions API: %w", err),
```
jayfreestone, line 188: `errors.Wrapf`

dimension/client.go
```diff
@@ -100,37 +141,76 @@ func (c *Client) GetAreaTypes(ctx context.Context, userAuthToken, serviceAuthTok
...
+		"dataset_id":   datasetID,
+		"area_type_id": areaTypeID,
+		"text":         text,
+	}
+
+	urlPath := "areas"
+	urlValues := url.Values{"dataset": []string{datasetID}, "area-type": []string{areaTypeID}}
+	if text != "" {
+		urlValues.Add("text", text)
+	}
```
jayfreestone, line 172: Although an empty string should still be a no-op, it's an optional field so might be worth having the same treatment as `text`.

dimension/client.go
```diff
@@ -100,37 +141,76 @@ func (c *Client) GetAreaTypes(ctx context.Context, userAuthToken, serviceAuthTok
...
+
+	req, err := c.createDimensionGetRequest(ctx, userAuthToken, serviceAuthToken, urlPath, logData, urlValues)
+	if err != nil {
+		return GetAreasResponse{}, dperrors.New(
+			err,
+			http.StatusBadRequest,
+			logData,
+		)
+	}
```
jayfreestone, line 181: I _think_ by doing this you wipe out the status code provided by `createDimensionGetRequest` (i.e. a `500` for a garbage URL`). In order to return a status code from those, you'd need to probably pull the status code at the callsite?

```go
	req, err := c.createDimensionGetRequest(ctx, userAuthToken, serviceAuthToken, urlPath, logData, urlValues)
	if err != nil {
		return GetAreasResponse{}, dperrors.New(
			err,
			dperrors.StatusCode(err),
			logData,
		)
	}
```

Might be wrong but worth checking. If you do change it, then you probably don't need to pass `logData` down to `createDimensionGetRequest` either since it doesn't add anything to it.

dimension/areas.go
```diff
@@ -0,0 +1,13 @@
+package dimension
+
+// Area is an area model with ID and Label
+type Area struct {
+	ID       string `json:"id"`
+	Label    string `json:"label"`
+	AreaType string `json:"area-type"`
```
ChrysmOre, line 7: I think we use snake_case for json fields

dimension/client.go
```diff
@@ -55,36 +56,76 @@ func (c *Client) Checker(ctx context.Context, check *healthcheck.CheckState) err
...
+			errors.Wrap(err, "failed to create request"),
+			http.StatusBadRequest,
+			logData,
+		)
+	}
+	return req, nil
+}
+
+func checkDimensionGetResponse(logData log.Data, resp http.Response) error {
+	if resp.StatusCode == http.StatusNotFound {
```
ChrysmOre, line 84: Why is StatusNotFound special? Sure any error code can (should) return errors in the same format?

dimension/client.go
```diff
@@ -55,36 +56,76 @@ func (c *Client) Checker(ctx context.Context, check *healthcheck.CheckState) err
...
+		return &http.Request{}, dperrors.New(
+			errors.Wrap(err, "failed to create request"),
+			http.StatusBadRequest,
+			logData,
+		)
+	}
+	return req, nil
+}
+
+func checkDimensionGetResponse(logData log.Data, resp http.Response) error {
```
ChrysmOre, line 83: You can drop the `dimension` from this name as it's already in package `dimension`. checkGetResponse would be clearer (and the same for the createDimensionGetRequest function too)

ChrysmOre, line 83: No need to pass `logData` around in these functions. So long as you return logData embedded in the errors and errors are wrapped at each level the data for all errors will be unwrapped at the same time

dimension/client.go
```diff
@@ -55,36 +56,76 @@ func (c *Client) Checker(ctx context.Context, check *healthcheck.CheckState) err
...
 	}
 
-	areaTypesURL.RawQuery = url.Values{"dataset": []string{datasetID}}.Encode()
-	reqURL := areaTypesURL.String()
+	return nil
+}
 
-	clientlog.Do(ctx, "getting area types", service, reqURL, logData)
+// GetAreaTypes retrieves the Cantabular area-types associated with a dataset
+func (c *Client) GetAreaTypes(ctx context.Context, userAuthToken, serviceAuthToken, datasetID string) (GetAreaTypesResponse, error) {
```
ChrysmOre, line 109: I think it would be better to use an input struct for these functions rather than long lists of arguments
`GetAreaTypes(ctx context.Context, input GetAreaTypesInput)`
This way it will be easier to change in the future without breaking (and will be tidier too)

dimension/areas.go
jackrehaag: Here it seems to be kebab-case
https://github.com/ONSdigital/dp-cantabular-dimension-api/blob/cf38fd889eeffd2243149995ac0f61b9eeb5480d/contract/areas.go#L8

dimension/client.go
jackrehaag: From the existing tests it seems like this is to preserve the original error type from the response body.
https://github.com/ONSdigital/dp-api-clients-go/blob/main/dimension/client_test.go#L116

dimension/client.go
jackrehaag: I've made this change for `GetAreas` but `GetAreaTypes` was not added in this PR and is actually already used so changing it now would break the frontend.

PR #249, opened 2022-04-28
Add dp-download-service client

mryan321:
### What

Add dp-download-service client:
- follow pattern of other clients (lift n shift'ish 🤠)
- only one endpoint covered - as expected for this piece of work: https://github.com/ONSdigital/dp-download-service/blob/500edf0ca0b286f8542f4f7b831bd9168749ff86/swagger.yml#L72
- have tested locally

### How to review

Sanity check and check test scenarios.

### Who can review

Anyone

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -1,0 +1,127 @@
+package download
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/ONSdigital/dp-api-clients-go/v2/clientlog"
+	dperrors "github.com/ONSdigital/dp-api-clients-go/v2/errors"
+	healthcheck "github.com/ONSdigital/dp-api-clients-go/v2/health"
+	health "github.com/ONSdigital/dp-healthcheck/healthcheck"
+	dprequest "github.com/ONSdigital/dp-net/request"
+	"github.com/ONSdigital/log.go/v2/log"
+)
+
+const service = "download-service"
+
+type Response struct {
+	Content io.ReadCloser `json:"content"`
+}
+
+// Client is an download service client which can be used to make requests to the server.
+// It extends the generic healthcheck Client structure.
+type Client struct {
+	hcCli            *healthcheck.Client
+	serviceAuthToken string
+}
+
+// NewAPIClient creates a new instance of DownloadServiceAPI Client with a given download service url
+func NewAPIClient(downloadServiceAPIURL, serviceAuthToken string) *Client {
+	return &Client{
+		healthcheck.NewClient(service, downloadServiceAPIURL),
+		serviceAuthToken,
+	}
+}
+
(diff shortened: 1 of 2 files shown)


download/download.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return nil, dperrors.New(
+			fmt.Errorf("failed to create request to DownloadService API: %s", err),
+			http.StatusInternalServerError,
+			nil,
+		)
+	}
+	defer closeResponseBody(ctx, resp)
+
+	b, err := ioutil.ReadAll(resp.Body)
```
braddle, line 83: Doing this will load the entire file into memory at this point. Would you now be better returning resp.Body in the response object and then allow the user of the client library to make a decision on how they want to read through the content.

download/download.go
```diff
@@ -0,0 +1,119 @@
...
+
+	if resp.StatusCode == http.StatusMovedPermanently {
+		return &Response{RedirectUrl: resp.Header.Get("location")}, nil
+	} else if resp.StatusCode != http.StatusOK {
+		return nil, dperrors.New(
+			errors.New(string(b)), resp.StatusCode, nil,
+		)
+	}
+
+	return &Response{Content: ioutil.NopCloser(bytes.NewBuffer(b))}, nil
```
braddle, line 100: As suggested about make `Content` Resp.body

download/download.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return nil, dperrors.New(
+			fmt.Errorf("failed to read error response body: %s", err),
+			resp.StatusCode,
+			nil,
+		)
+	}
+
+	if resp.StatusCode == http.StatusMovedPermanently {
+		return &Response{RedirectUrl: resp.Header.Get("location")}, nil
```
braddle, line 93: Could you resolve the redirect and just return a ReaderCloser from this function rather than a Response object.

download/download_test.go
```diff
@@ -0,0 +1,336 @@
+package download
```
braddle, line 1: Can I recommend you look at the testing of the Files client https://github.com/ONSdigital/dp-api-clients-go/blob/main/files/client_test.go

We took a different approach to testing which enabled us to simplify the client interface buy removing a number of functions that only exist for testing like URL() and HealthClient()

braddle (changes requested):
I have added a few suggestions to the PR

download/download.go
mryan321: yep makes sense 👍

download/download.go
mryan321: 👍  i think good to know its a redirect though so am thinking to keep `RedirectUrl` in response - can just ignore. just thinking for us - we will persist this and stop the redirect next time. But we will prob persist on publish anyway so never hit this scenario 🙈

mryan321: forget this - httpclient by default follows redirect 😄

braddle (approved):
LGTM

PR #260, opened 2022-05-12
Updated ModelDimension

j-s-rawat:
### What

Added ID and Label to ModelDimension

### How to review
The Dimension data model in filter-flex had ID and Label which was missing in api-client side

### Who can review

Anyone


PauloPortugal (approved):
LGTM

PR #262, opened 2022-05-18
Expect HTTP Accepted code from  filters/{id}/submit

PauloPortugal:
### What

The **dp-cantabular-filter-flex-api** `/filter/{id}/submit` [endpoint](https://github.com/ONSdigital/dp-cantabular-filter-flex-api/blob/develop/api/filters.go#L221) is returning a `202 Accepted` HTTP response code.

### How to review

Ensure tests run and check the **dp-cantabular-filter-flex-api** project.

### Who can review

Any developer

### DIFF
--- a/filter/data.go
+++ b/filter/data.go
@@ -149,12 +149,9 @@ type SubmitFilterRequest struct {
 }
 
 type SubmitFilterResponse struct {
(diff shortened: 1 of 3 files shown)


saminahbab (approved):
LGTM.

PR #271, opened 2022-06-13
Feature/add release attributes

PatrickConnollyONS:
### What

Add WelshStatistic and Survey attributes to the Release type for zebedee and releasecalendar
Some small tidy ups

### How to review

Review code and ensure test are passing

### Who can review

Anyone

### DIFF
--- a/zebedee/data.go
+++ b/zebedee/data.go
@@ -90,6 +90,8 @@ type Description struct {
 	Keywords           []string `json:"keywords"`
 	MetaDescription    string   `json:"metaDescription"`
 	NationalStatistic  bool     `json:"nationalStatistic"`
+	WelshStatistic     bool     `json:"welshStatistic"`
+	Survey             string   `json:"survey"`
 	LatestRelease      bool     `json:"latestRelease"`
 	Contact            Contact  `json:"contact"`
 	ReleaseDate        string   `json:"releaseDate"`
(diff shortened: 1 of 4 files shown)


zebedee/client.go
```diff
@@ -192,7 +182,7 @@ func (c *Client) put(ctx context.Context, userAccessToken, path string, payload
 }
 
 // GetBreadcrumb returns a Breadcrumb
-func (c *Client) GetBreadcrumb(ctx context.Context, userAccessToken, collectionID, lang, uri string) ([]Breadcrumb, error) {
```
rafahop, line 195: If we do not need these params, should we remove them?

zebedee/data.go
```diff
@@ -134,10 +136,6 @@ type Alert struct {
 	Type     string `json:"type"`
 }
 
-// Related stores the Title and URI for any related data (e.g. related publications on a dataset page)
-// Deprecated: use Link
-type Related = Link
```
rafahop, line 139: Is this a breaking change?

PatrickConnollyONS: yes it is - I have removed it

zebedee/client.go
PatrickConnollyONS: yes I think so. However I have reverted most of these 'tidy-ups' as they are introducing breaking changes

PR #281, opened 2022-06-30
add cantabluar get parents func

ChrysmOre:
### What

Added Cantabular client function to return parents of a given variable.

Also removed redundant separation of DatasetVariables & DatasetRuleBase to be one struct, Dataset

### How to review

Check changes make sense, tests pass

### Who can review

Anyone

### DIFF
--- a/cantabular/contract.go
+++ b/cantabular/contract.go
@@ -71,7 +71,7 @@ type SearchDimensionsRequest struct {
 // POST [cantabular-ext]/graphql
 // with a query to obtain variables
 type GetDimensionsResponse struct {
-	Dataset gql.DatasetVariables `json:"dataset"`
+	Dataset gql.Dataset `json:"dataset"`
 }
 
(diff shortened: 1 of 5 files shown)


cantabular/contract.go
```diff
@@ -71,7 +71,7 @@ type SearchDimensionsRequest struct {
 // POST [cantabular-ext]/graphql
 // with a query to obtain variables
 type GetDimensionsResponse struct {
-	Dataset gql.DatasetVariables `json:"dataset"`
+	Dataset gql.Dataset `json:"dataset"`
```
PauloPortugal, line 74: Was this incorrect?

ChrysmOre: It was needlessly split into two different structs, where in reality it's the same struct that can have one of two fields returned

PauloPortugal (approved):

PR #282, opened 2022-06-30
5732 - Add delete dimension options function

PauloPortugal:
### What

`DeleteDimensionOptions` completely removes the options array from a
given dimension and filterID

Resolves: [5732 - dp-api-clients-go function for DELETE filters/{filterid}/dimensions/{dimension}/options](https://trello.com/c/XBSVWDXW/5732-dp-api-clients-go-function-for-delete-filter-filterid-dimension-dimension-options)

### How to review

Sense check it.

### Who can review

Any ONS developer.

### DIFF
--- a/filter/filter.go
+++ b/filter/filter.go
@@ -10,6 +10,8 @@ import (
 	"net/http"
 	"strconv"
 
+	"github.com/pkg/errors"
+
 	"github.com/ONSdigital/dp-api-clients-go/v2/batch"
 	"github.com/ONSdigital/dp-api-clients-go/v2/clientlog"
 	dperrors "github.com/ONSdigital/dp-api-clients-go/v2/errors"
@@ -18,7 +20,6 @@ import (
 	health "github.com/ONSdigital/dp-healthcheck/healthcheck"
 	dprequest "github.com/ONSdigital/dp-net/request"
 	"github.com/ONSdigital/log.go/v2/log"
-	"github.com/pkg/errors"
 )
 
 const service = "filter-api"
@@ -476,6 +477,42 @@ func (c *Client) GetDimensionOptionsBatchProcess(ctx context.Context, userAuthTo
 	}
 
 	return eTag, batch.ProcessInConcurrentBatches(batchGetter, batchProcessor, batchSize, maxWorkers)
+}
+
+// DeleteDimensionOptions completely removes the options array from a given dimension
+func (c *Client) DeleteDimensionOptions(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, filterID, name string) (string, error) {
+	logData := log.Data{
+		"filter_id":      filterID,
+		"dimension_name": name,
+	}
+
+	uri := fmt.Sprintf("%s/filters/%s/dimensions/%s/options", c.hcCli.URL, filterID, name)
+	clientlog.Do(ctx, "deleting selected dimension options", service, uri)
+
+	res, err := c.doDeleteWithAuthHeadersAndWithDownloadToken(ctx, userAuthToken, serviceAuthToken, collectionID, uri)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to delete dimension options")
+	}
+
+	defer closeResponseBody(ctx, res)
+
+	if res.StatusCode != http.StatusNoContent {
+		return "", dperrors.New(
+			errors.Wrap(&ErrInvalidFilterAPIResponse{http.StatusNoContent, res.StatusCode, uri}, "unexpected response"),
+			res.StatusCode,
+			logData,
+		)
+	}
+
+	eTag, err := headers.GetResponseETag(res)
+	if err != nil && err != headers.ErrHeaderNotFound {
+		dperrors.New(
+			errors.Wrap(err, "no ETag header found"),
+			res.StatusCode,
+			logData,
+		)
+	}
+	return eTag, err
 }
 
 // CreateFlexibleBlueprint creates a flexible filter blueprint and returns the associated filterID and eTag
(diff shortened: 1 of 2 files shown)


filter/filter.go
```diff
@@ -478,6 +479,33 @@ func (c *Client) GetDimensionOptionsBatchProcess(ctx context.Context, userAuthTo
 	return eTag, batch.ProcessInConcurrentBatches(batchGetter, batchProcessor, batchSize, maxWorkers)
 }
 
+// DeleteDimensionOptions completely removes the options array from a given dimension
+func (c *Client) DeleteDimensionOptions(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, filterID, name string) (eTag string, err error) {
+
```
ChrysmOre, line 484: No need for blank line at beginning of functions

filter/filter.go
```diff
@@ -478,6 +479,33 @@ func (c *Client) GetDimensionOptionsBatchProcess(ctx context.Context, userAuthTo
 	return eTag, batch.ProcessInConcurrentBatches(batchGetter, batchProcessor, batchSize, maxWorkers)
 }
 
+// DeleteDimensionOptions completely removes the options array from a given dimension
```
ChrysmOre, line 482: Can we not use named return arguments unless there's a need to (manipulating return arguments in deferred functions)?

filter/filter.go
```diff
@@ -478,6 +479,33 @@ func (c *Client) GetDimensionOptionsBatchProcess(ctx context.Context, userAuthTo
...
+func (c *Client) DeleteDimensionOptions(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, filterID, name string) (eTag string, err error) {
+
+	uri := fmt.Sprintf("%s/filters/%s/dimensions/%s/options", c.hcCli.URL, filterID, name)
+	clientlog.Do(ctx, "deleting selected dimension options", service, uri)
+
+	resp, err := c.doDeleteWithAuthHeadersAndWithDownloadToken(ctx, userAuthToken, serviceAuthToken, collectionID, uri)
+
+	if err != nil {
```
ChrysmOre, line 490: Please wrap errors (and remove this blank line between checking the error)

filter/filter.go
```diff
@@ -478,6 +479,33 @@ func (c *Client) GetDimensionOptionsBatchProcess(ctx context.Context, userAuthTo
...
+	}
+
+	defer closeResponseBody(ctx, resp)
+
+	if resp.StatusCode != http.StatusNoContent {
+		err = &ErrInvalidFilterAPIResponse{http.StatusNoContent, resp.StatusCode, uri}
+		return "", err
+	}
+
+	eTag, err = headers.GetResponseETag(resp)
```
ChrysmOre, line 501: Is this forcing us to send an eTag? We need to be sure it's ok to omit it and it will be treated as '*'

filter/filter.go
```diff
@@ -478,6 +479,33 @@ func (c *Client) GetDimensionOptionsBatchProcess(ctx context.Context, userAuthTo
...
+	resp, err := c.doDeleteWithAuthHeadersAndWithDownloadToken(ctx, userAuthToken, serviceAuthToken, collectionID, uri)
+
+	if err != nil {
+		return "", err
+	}
+
+	defer closeResponseBody(ctx, resp)
+
+	if resp.StatusCode != http.StatusNoContent {
+		err = &ErrInvalidFilterAPIResponse{http.StatusNoContent, resp.StatusCode, uri}
```
ChrysmOre, line 497: Why no just `return &ErrInvalid`...?

filter/filter.go
PauloPortugal: :facepalm: This has been addressed.

filter/filter.go
PauloPortugal: Another copy and paste... 
Let me change that.

filter/filter.go
PauloPortugal: This has been addressed.

filter/filter.go
PauloPortugal: This has been addressed

filter/filter.go
PauloPortugal: Done.

filter/filter.go
PauloPortugal: This has been addressed

filter/filter.go
```diff
@@ -478,6 +479,29 @@ func (c *Client) GetDimensionOptionsBatchProcess(ctx context.Context, userAuthTo
...
+
+	resp, err := c.doDeleteWithAuthHeadersAndWithDownloadToken(ctx, userAuthToken, serviceAuthToken, collectionID, uri)
+	if err != nil {
+		return "", errors.Wrap(err, "an error occurred when attempting to delete dimension options")
+	}
+
+	defer closeResponseBody(ctx, resp)
+
+	if resp.StatusCode != http.StatusNoContent {
+		return "", &ErrInvalidFilterAPIResponse{http.StatusNoContent, resp.StatusCode, uri}
```
ChrysmOre, line 495: Sorry I didn't see this before but I don't like this, we're dropping the error response from the service entirely. I know this is how this client is already designed but it's designed badly. Can we work out a way to return the response body returned in these cases? I think in the response body would be preferable but at the very least in some logData?

filter/filter.go
```diff
@@ -478,6 +479,29 @@ func (c *Client) GetDimensionOptionsBatchProcess(ctx context.Context, userAuthTo
...
 }
 
+// DeleteDimensionOptions completely removes the options array from a given dimension
+func (c *Client) DeleteDimensionOptions(ctx context.Context, userAuthToken, serviceAuthToken, collectionID, filterID, name string) (string, error) {
+	uri := fmt.Sprintf("%s/filters/%s/dimensions/%s/options", c.hcCli.URL, filterID, name)
+	clientlog.Do(ctx, "deleting selected dimension options", service, uri)
+
+	resp, err := c.doDeleteWithAuthHeadersAndWithDownloadToken(ctx, userAuthToken, serviceAuthToken, collectionID, uri)
+	if err != nil {
+		return "", errors.Wrap(err, "an error occurred when attempting to delete dimension options")
```
ChrysmOre, line 489: I won't say this is a blocker this time but in general I think error messages should be as concise as possible. `failed to` is preferable to `an error occurred when attempting to` especially as these errors generally get stitched together to form long chains, things can quickly get overly convoluted.

PauloPortugal: This has been addressed.

filter/filter.go
PauloPortugal: I have returned the `logData`

saminahbab (approved):

saminahbab (approved):

PR #283, opened 2022-07-01
Added support for delete options 

j-s-rawat:
### What

Added support for filters/{filterid}/dimensions/{dimension}/options/{option}

### How to review

- The interface is returning the right set of information 
- Right endpoint is being called
- All tests are passing

### Who can review

Anyone


filterflex/client.go
```diff
@@ -70,3 +75,106 @@ func (c *Client) ForwardRequest(req *http.Request) (*http.Response, error) {
...
+	if err != nil {
+		return "", dperrors.New(
+			errors.Wrap(err, "failed to get response from filter flex API"),
+			http.StatusInternalServerError,
+			logData,
+		)
+	}
+	defer closeResponseBody(ctx, resp)
+
+	if resp.StatusCode != http.StatusOK {
```
PauloPortugal, line 157: I believe we should be matching against `204` and not `200`
https://github.com/ONSdigital/dp-cantabular-filter-flex-api/blob/develop/api/filter_dimension_options.go#L270

PauloPortugal (commented):
Please align the HTTP response expectation with the filter-flex-api response code.
I did a similar thing on another PR :slightly_smiling_face: 
Other than that looks really good :+1:

filterflex/client.go
j-s-rawat: will update

PauloPortugal (dismissed):
LGTM

PR #286, opened 2022-07-13
Added support for areaTypesPatents

j-s-rawat:
### What

Introduced API client function to support getting parents for areaTypes

### How to review

- All test pass 
- The response matches the end-point in dp-population-types-api

### Who can review

Anyone

### DIFF
--- a/population/area_types.go
+++ b/population/area_types.go
@@ -1,6 +1,6 @@
 package population
 
-// AreaType is an area-type model with ID and Label
+// AreaType is an area type model with ID and Label
 type AreaType struct {
 	ID         string `json:"id"`
 	Label      string `json:"label"`
@@ -9,5 +9,5 @@ type AreaType struct {
 
 // GetAreaTypesResponse is the response object for GET /area-types
 type GetAreaTypesResponse struct {
-	AreaTypes []AreaType `json:"area-types"`
+	AreaTypes []AreaType `json:"area_types"`
 }
--- a/population/areas.go
+++ b/population/areas.go
@@ -4,10 +4,22 @@ package population
 type Area struct {
 	ID       string `json:"id"`
 	Label    string `json:"label"`
-	AreaType string `json:"area-type"`
+	AreaType string `json:"area_type"`
 }
 
 // GetAreasResponse is the response object for GET /areas
 type GetAreasResponse struct {
 	Areas []Area `json:"areas"`
+}
+
+// Area is an area model with ID and Label
+type AreaTypes struct {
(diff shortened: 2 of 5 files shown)


nimitsarup (dismissed):

population/area_types.go
```diff
@@ -1,13 +1,13 @@
...
-// AreaType is an area-type model with ID and Label
+// AreaType is an area type model with ID and Label
 type AreaType struct {
 	ID         string `json:"id"`
 	Label      string `json:"label"`
 	TotalCount int    `json:"total_count"`
 }
 
-// GetAreaTypesResponse is the response object for GET /area-types
+// GetAreaTypesResponse is the response object for GET /area_types
```
ChrysmOre, line 10: URLs are `camel-case`, JSON is `snake_case`

population/areas.go
```diff
@@ -4,10 +4,22 @@ package population
...
 // GetAreasResponse is the response object for GET /areas
 type GetAreasResponse struct {
 	Areas []Area `json:"areas"`
 }
+
+// Area is an area model with ID and Label
+type AreaTypes struct {
+	ID         string `json:"id"`
+	Label      string `json:"label"`
+	TotalCount string `json:"total-count"`
```
ChrysmOre, line 19: you probable wanted to change to `total_count` here

population/client_test.go
```diff
@@ -530,6 +530,171 @@ func TestGetPopulationTypes(t *testing.T) {
...
+			UserAuthToken:    userAuthToken,
+			ServiceAuthToken: serviceAuthToken,
+		}
+		_, err := client.GetAreaTypesParents(context.Background(), input)
+
+		Convey("it should return an internal error", func() {
+			So(err, shouldBeDPError, http.StatusInternalServerError)
+		})
+	})
+	/* Currently the swagger spec in dp-population-type-api doesn't expect any error codes
```
ChrysmOre, line 618: You should probably uncomment these tests or remove them. The only hard-coded status code in the endpoint is 500 but any status code returned from Cantabular can also be returned. And there are many reasons a status code could be returned other than what is defined in the swagger spec, if Cantabular gets updated or the service is unreachable etc

population/areas.go
ChrysmOre: Also shouldn't AreaTypes be in area_type.go? Same for GetAreaTypesParentsResponse. Actually we should have a contract.go for that..

PauloPortugal (approved):
:+1:

PR #290, opened 2022-07-15
Update Cantabular parent query

ChrysmOre:
### What

Updated Cantabular parents query to return all parents, not just direct parent

### How to review

Check change makes sense

### Who can review

Anyone

### DIFF
--- a/cantabular/dimensions.go
+++ b/cantabular/dimensions.go
@@ -213,5 +213,14 @@ func (c *Client) GetParents(ctx context.Context, req GetParentsRequest) (*GetPar
 		)
 	}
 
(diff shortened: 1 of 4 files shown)


redhug1 (dismissed):
Having read the thread between you and sensible code, i can see you have sliced off the last element with:
[:len(edges)-1]
So, LGTM

redhug1 (commented):
When i pressed the Compare button in github for your re-base action and looked through that list of changes, i can see line 279 in cantabular/queries.go should probably not be there ?
its:
    // return err

redhug1 (approved):
LGTM

PR #297, opened 2022-07-25
Update Geography query to use rule: true instead of ruleBase

ChrysmOre:
### What

Updated QueryGeographyDimensions to use variables(rule: true) instead of ruleBase as the latter is now deprecated.

### How to review

Check changes make sense, tests pass

### Who can review

anyone


PauloPortugal (approved):
LGTM

PR #300, opened 2022-08-01
feat: (pop-types-api) get area method

saminahbab:
### What

Trello #5796 add method for
GET population-types/{type}/areas/{type}/area/{id}

### How to review

run make test
ensure that the body of the api method is up to spec 

### Who can review

Team B

### DIFF
--- a/population/client.go
+++ b/population/client.go
@@ -27,6 +27,14 @@ type Client struct {
 	baseURL *url.URL
 }
 
+type GetAreaInput struct {
+	UserAuthToken    string
+	ServiceAuthToken string
+	PopulationType   string
+	AreaType         string
+	Area             string
+}
+
 type GetAreasInput struct {
 	UserAuthToken    string
 	ServiceAuthToken string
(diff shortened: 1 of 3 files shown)


population/client.go
```diff
@@ -271,6 +279,59 @@ func (c *Client) GetAreas(ctx context.Context, input GetAreasInput) (GetAreasRes
 	return areas, nil
 }
 
+func (c *Client) GetArea(ctx context.Context, input GetAreaInput) (Area, error) {
+
```
ChrysmOre, line 283: No need for blank lines at the beginning and end of functions

population/client.go
```diff
@@ -271,6 +279,59 @@ func (c *Client) GetAreas(ctx context.Context, input GetAreasInput) (GetAreasRes
...
+	logData := log.Data{
+		"method":          http.MethodGet,
+		"population_type": input.PopulationType,
+		"area_type":       input.AreaType,
+		"area_type_id":    input.AreaTypeID,
+	}
+	urlPath := fmt.Sprintf("population-types/%s/area-types/%s/areas/%s", input.PopulationType, input.AreaType, input.AreaTypeID)
```
ChrysmOre, line 290: Why `urlPath` instead of just `url?`

population/client.go
```diff
@@ -271,6 +279,59 @@ func (c *Client) GetAreas(ctx context.Context, input GetAreasInput) (GetAreasRes
 	return areas, nil
 }
 
+func (c *Client) GetArea(ctx context.Context, input GetAreaInput) (Area, error) {
```
ChrysmOre, line 282: Looks like the existing functions follow the pattern of returning fooResponse with embedded foo rather than just raw foo

population/client.go
saminahbab: its the standard set in the other methods in that file.

`func (c *Client) GetAreas(ctx context.Context, input GetAreasInput) (GetAreasResponse, error) {
	logData := log.Data{
		"method":       http.MethodGet,
		"dataset_id":   input.DatasetID,
		"area_type_id": input.AreaTypeID,
		"text":         input.Text,
	}

	urlPath := fmt.Sprintf("population-types/%s/area-types/%s/areas", input.DatasetID, input.AreaTypeID)`

for example.

ChrysmOre: Fair enough, I'd say that elsewhere it's a slightly bad choice too. I guess it's not a huge thing but it's a Go paradigm to choose shorter variable names wherever possible, you'll find that other organisations will be much more attentive to these kind of details than we tend to be here. It's a good habit to learn the idioms and preferred styles of the language. https://github.com/golang/go/wiki/CodeReviewComments#variable-names

population/client.go
```diff
@@ -271,6 +279,58 @@ func (c *Client) GetAreas(ctx context.Context, input GetAreasInput) (GetAreasRes
...
+	if err := json.NewDecoder(resp.Body).Decode(&area); err != nil {
+		return GetAreaResponse{}, dperrors.New(
+			errors.Wrap(err, "unable to deserialize areas response"),
+			http.StatusInternalServerError,
+			logData,
+		)
+	}
+
+	return GetAreaResponse{Area: area}, nil
+
```
ChrysmOre, line 331: blank line

population/client.go
saminahbab: No im totally okay with what is there already. 
There is a url package that is quite popular that is probably going to clash with your autosuggest if you try to use that as a variable name, same with path btw. 
I'd rather keep `urlPath`

saminahbab: will definitely read through the best practice link though for future!

population/client.go
```diff
@@ -27,6 +27,14 @@ type Client struct {
 	baseURL *url.URL
 }
 
+type GetAreaInput struct {
+	UserAuthToken    string
+	ServiceAuthToken string
+	PopulationType   string
+	AreaType         string
+	AreaTypeID       string
```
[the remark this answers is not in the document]
saminahbab: Changed!

andre-urbani (approved):
LGTM

PR #305, opened 2022-08-23
Feature/cantabular/add new get dimensions query

ChrysmOre:
### What

Added new GetDimensions function which retrieves all non-geography dimensions, renamed existing GetDimensions (which isn't used anywhere) to GetAllDimensions
### How to review

Check changes make sense

### Who can review

Anyone

### DIFF
--- a/cantabular/dimensions.go
+++ b/cantabular/dimensions.go
@@ -14,7 +14,7 @@ import (
 // GetGeographyBatchProcessor is the type corresponding to a batch processing function for Geography dimensions
 type GetGeographyBatchProcessor func(response *GetGeographyDimensionsResponse) (abort bool, err error)
 
-// GetDimensions performs a graphQL query to obtain all the dimensions for the provided cantabular dataset.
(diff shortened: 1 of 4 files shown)


cantabular/dimensions.go
```diff
@@ -26,10 +26,45 @@ func (c *Client) GetDimensions(ctx context.Context, dataset string) (*GetDimensi
...
+			resp.Errors[0].StatusCode(),
+			log.Data{"errors": resp.Errors},
+		)
+	}
+
+	return &resp.Data, nil
+}
+
+// GetDimensions performs a graphQL query to obtain all the non-geography dimensions for the provided
+// cantabular dataset. The whole response is loaded to memory.
```
saminahbab, line 45: I suppose this is the only way? 
Are we aware of how big the responses can be?

ChrysmOre: The size of the responses for lists of dimensions won't be significant enough to worry about memory, that's only really relevant for options and counts

ChrysmOre: That comment is just a slightly modified version of what was there before

saminahbab (approved):

PR #306, opened 2022-08-23
5829 - Updates to columns in CSV Download

PauloPortugal:
### What

A new column is required in the CSV file that is created for the Cantabular downloads. This should hold the category code for the row that the observation is being returned for.

* [5829 - Move "Count" column to the end of row](https://github.com/ONSdigital/dp-api-clients-go/commit/2b88db3115413ffa1aab5d2ea7230dcc590e2cb0)
Rather than the beginning of the row. The word `Count` should also be renamed to `Observation`

* [5829 - Add Category Code column for each Category Label](https://github.com/ONSdigital/dp-api-clients-go/commit/f33d77900bd966f4de81f76eb8aff0d7f67f9b39) 
A new column with the same Label string appended with `Code` should precede it's respective `Label` column.

**Resolves:** [5829](https://trello.com/c/N8L65t3B/5829-updates-to-columns-in-csv-download)

### How to review

1. Ensure tests pass
2. Sense check it
3. Check the story requirements

### Who can review

Any ONS developer


cantabular/table.go
```diff
@@ -298,11 +298,12 @@ func decodeValues(ctx context.Context, dec jsonstream.Decoder, dims Dimensions,
...
 func createCSVHeader(dims Dimensions) []string {
-	header := make([]string, len(dims)+1)
+	l := len(dims)
+	header := make([]string, l+1)
 	for i, dim := range dims {
-		header[i+1] = dim.Variable.Label
+		header[i] = dim.Variable.Label
 	}
-	header[0] = "count"
+	header[l] = "Observation"
```
thomasridd, line 310: Query: Should this be "Observation" or "Observations"
Ticket says "Observation" but I'd just like it checked

cantabular/table.go
```diff
@@ -298,11 +298,16 @@ func decodeValues(ctx context.Context, dec jsonstream.Decoder, dims Dimensions,
 // createCSVHeader creates an array of strings corresponding to a csv header
 // where each column contains the value of the corresponding dimension, with the last column being the 'count'
 func createCSVHeader(dims Dimensions) []string {
-	header := make([]string, len(dims)+1)
-	for i, dim := range dims {
-		header[i+1] = dim.Variable.Label
+	i := 0
+	l := len(dims) * 2
+	header := make([]string, l+1)
+	for _, dim := range dims {
```
ChrysmOre, line 304: Could you initialise `i` here in the loop declaration and remove the second incrementation?

cantabular/table.go
thomasridd: All good - ignore me

cantabular/table.go
PauloPortugal: I am not sure this would be a better option

PauloPortugal: I would still stick with my initial approach, as it is easier to read the code.

j-s-rawat: I am approving this as either approach are acceptable and won't effect the outcome

j-s-rawat (approved):
LGTM

PR #313, opened 2022-08-31
Cantabular - unmarshal into correct GetAreaResponse object

ChrysmOre:
### What

Fix bug where Cantabular/GetArea would be responding with an empty struct as it was attempting to unmarshal the response from the service into an incorrect object.
There was not an error being returned because the response would be valid JSON, just the provided fields where incorrect.

### How to review

Check the changes make sense. 

### Who can review

Anyone


saminahbab (approved):
LGTM

PR #315, opened 2022-09-02
5802 Fix GetAreas 'totalCount' and 'count'

PauloPortugal:
### What

Resolves: [5802](https://trello.com/c/7MD0oUPR/5802-bug-fix-count-on-review-page)

### How to review

Sense check it and tests are green

### Who can review

Any ONS developer

### DIFF
--- a/cantabular/queries.go
+++ b/cantabular/queries.go
@@ -208,6 +208,7 @@ query ($dataset: String!, $text: String!, $category: String!, $limit: Int!, $off
 			name
 			label
 			categories {
+                          totalCount
 			  search(text: $category, first: $limit, skip: $offset ) {
 				edges {
 				  node {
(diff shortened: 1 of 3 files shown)


cantabular/queries.go
```diff
@@ -208,6 +208,7 @@ query ($dataset: String!, $text: String!, $category: String!, $limit: Int!, $off
 			name
 			label
 			categories {
+              totalCount
```
j-s-rawat, line 211: not a issue but we might indent it

j-s-rawat (dismissed):
LGTM

cantabular/queries.go
PauloPortugal: oh yes...

PauloPortugal: this has been addressed

j-s-rawat (approved):
LGTM

PR #317, opened 2022-09-05
Add RelatedAPIDatasets

PatrickConnollyONS:
### What

Add RelatedAPIDatasets to the Release objects in zebedee and releasecalendar packages

### How to review

Review code and ensure all tests pass

### Who can review

Anyone


zebedee/client_test.go
```diff
@@ -105,7 +105,7 @@ func contentData(w http.ResponseWriter, req *http.Request) {
...
+		w.Write([]byte(`{"markdown":["markdown"],"relatedDocuments":[{"uri":"pageDescription2"}],"relatedDatasets":[{"uri":"pageDescription1"}],"relatedAPIDatasets":[{"uri":"pageDescription2"}],"relatedMethodology":[{"uri":"pageDescription1"}],"relatedMethodologyArticle":[{"uri":"pageDescription2"}],"links":[{"uri":"pageDescription1"}, {"uri":"pageDescription2"}],"dateChanges":[{"previousDate":"2021-08-15T11:12:05.592Z","changeNotice":"change notice"}],"uri":"/releases/indexofproductionukdecember2021timeseries","description":{"finalised":true,"title":"Index of Production","summary":"Movements in the volume of production for the UK production industries","nationalStatistic":true,"contact":{"email":"indexofproduction@ons.gov.uk","name":"Contact name","telephone":"+44 1633 456980"},"releaseDate":"2022-02-11T07:00:00.000Z","nextRelease":"11 March 2022","cancelled":true,"cancellationNotice":["notice"],"finalised":true,"published":true,"provisionalDate":"Dec 22"}}`))
```
rafahop, line 108: I would use an unique uri here instead of `pageDescription2` so that we can be sure we are using the right value?

releasecalendar/client_test.go
```diff
@@ -177,7 +189,7 @@ func TestGetLegacyRelease(t *testing.T) {
 		})
 	})
 
-	Convey("Given that 200 OK is returned by the API with an empty body", t, func() {
+	Convey("Given that 500 StatusInternalServerError is returned by the API with an empty body", t, func() {
```
rafahop, line 192: Is this change correct? It looks like the test client does return 200 OK?

releasecalendar/client_test.go
```diff
@@ -142,7 +154,7 @@ func TestGetLegacyRelease(t *testing.T) {
 		})
 	})
 
-	Convey("Given that 200 OK is returned by the API with an invalid body", t, func() {
+	Convey("Given that 500 StatusInternalServerError is returned by the API with an invalid body", t, func() {
```
rafahop, line 157: Is this change correct? It looks like the test client does return 200 OK?

releasecalendar/client_test.go
PatrickConnollyONS: let me have a look

releasecalendar/client_test.go
PatrickConnollyONS: yep sorry I got confused as to who was returning the error

releasecalendar/client_test.go
PatrickConnollyONS: as below

rafahop (approved):

PR #318, opened 2022-09-06
create cantabular metadata client and GetDefaultClassification func

ChrysmOre:
### What

Added cantabular metadata client

Added first func, GetDefaultClassification

### How to review

Check changes make sense

### Who can review

Anyone

### DIFF
--- a/cantabularmetadata/client.go
+++ b/cantabularmetadata/client.go
@@ -1,0 +1,193 @@
+package cantabularmetadata
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	dperrors "github.com/ONSdigital/dp-api-clients-go/v2/errors"
+	"github.com/ONSdigital/dp-api-clients-go/v2/health"
+	"github.com/ONSdigital/dp-healthcheck/healthcheck"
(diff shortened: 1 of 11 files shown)


cantabularmetadata/client.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Client is the client for interacting with the Cantabular API
+type Client struct {
+	ua      httpClient
+	host    string
+	version string
+}
+
+// NewClient returns a new Client
+func NewClient(cfg Config, ua httpClient) *Client {
```
saminahbab, line 33: heads up... working on a ticket that needs this and realised that the client initialisation isnt consistent across all clients.... for example population types 
```
// NewClient creates a new instance of Client with a given Population Type API URL
func NewClient(apiURL string) (*Client, error) {
	client := health.NewClient(service, apiURL)
	return NewWithHealthClient(client)
}
```
this isnt consistent and is yielding jilted looking code

ChrysmOre: What's the issue? I don't think there's 100% line for line consistency across the clients either way. It's not something I'd worry about

saminahbab: just thought i would bring it up, cause we do try to aim for consistency, not a blocker

saminahbab (approved):

PR #319, opened 2022-09-06
5849 Create a 'GetArea' function to perform an exact match

PauloPortugal:
### What

Previously the endpoint
```
GET /population-types/{dataset}/area-types/{area-type}/areas/{area-id}
```

was performing a fuzzy search and limiting the results to 1.
Since cantabular api extension  v10.2.1 we can now match exactly an area id.

This is the change to perform an exact match.

Resolves: [5849](https://trello.com/c/epE8FvNN/5849-getarea-endpoint-returning-error-results)

### How to review

Sense check it and ensure tests are :green_circle: 

### Who can review

ONS developers

### DIFF
--- a/cantabular/contract.go
+++ b/cantabular/contract.go
@@ -137,11 +137,23 @@ type GetAreasRequest struct {
 	Category string
 }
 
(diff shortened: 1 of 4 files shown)


j-s-rawat (approved):
LGTM

PR #324, opened 2022-09-14
5858 Create 'GET' parent areas count function

PauloPortugal:
### What

A new GET endpoint is needed to retrieve the count of areas within the larger area-type.

URI would be
```
  /population-types/{population-type}/area-type/{area-type}/parents/{parent-area-type}/areas-count?areas=put,your,list,of,areas,here.
```

This should just return an integer of the count of the number of areas within the original area type within the larger area.

Resolves: [5858](https://trello.com/c/oDyIu8hU/5858-create-endpoint-to-get-the-count-of-areas-within-the-larger-area-type)
Depends on:  https://github.com/ONSdigital/dp-population-types-api/pull/75

### How to review

* Sense check it
* Ensure tests are :green_circle:

### Who can review

Any ONS developer

### DIFF
--- a/population/client.go
+++ b/population/client.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -48,6 +49,15 @@ type GetAreaTypeParentsInput struct {
 	ServiceAuthToken string
 	DatasetID        string
 	AreaTypeID       string
+}
+
+type GetParentAreaCountInput struct {
+	UserAuthToken    string
+	ServiceAuthToken string
+	DatasetID        string
+	AreaTypeID       string
+	ParentAreaTypeID string
+	Areas            []string
 }
 
 type GetDimensionsInput struct {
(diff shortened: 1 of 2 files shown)


population/client.go
```diff
@@ -50,6 +51,15 @@ type GetAreaTypeParentsInput struct {
 	AreaTypeID       string
 }
 
+type GetParentsAreaCountInput struct {
```
ChrysmOre, line 54: Nit pick but should be Parent singular

population/client.go
```diff
@@ -405,6 +415,70 @@ func (c *Client) GetAreaTypeParents(ctx context.Context, input GetAreaTypeParent
...
+	logData := log.Data{
+		"method":              http.MethodGet,
+		"dataset_id":          input.DatasetID,
+		"area_type_id":        input.AreaTypeID,
+		"parent_area_type_id": input.ParentAreaTypeID,
+		"areas":               input.Areas,
+	}
+
+	urlPath := fmt.Sprintf("population-types/%s/area-types/%s/parents/%s/areas-count",
+		input.DatasetID, input.AreaTypeID, input.ParentAreaTypeID)
```
ChrysmOre, line 428: More Go-typical formatting might look something like:

```
urlPath := fmt.Sprintf(
  "population-types/%s/area-types/%s/parents/%s/areas-count",
  input.DatasetID,
  input.AreaTypeID,
  input.ParentAreaTypeID,
)
```
Certainly close the bracket on it's own line

PauloPortugal: This has been addressed.

population/client.go
PauloPortugal: This has been addressed.

andre-urbani (approved):
LGTM

PR #327, opened 2022-09-20
5858 - Amend the parent areas count type

PauloPortugal:
### What

* The value returned by the Populations API is an `int` and not a `string`.
* Amend the parent areas count query param structure  
    It should be `?areas=area1,area2`
    And not `?areas=area1&areas=area2`

**Resolves**: [5843](https://trello.com/c/cKGTgs1o/5843-fix-count-on-pages-that-getareas) [5858](https://trello.com/c/oDyIu8hU/5858-create-endpoint-to-get-the-count-of-areas-within-the-larger-area-type)

### How to review

* Sense check it
* Ensure test are :green_circle: 
* Check with @mr-nick17 

### Who can review

Any ONS developer

### DIFF
--- a/population/client.go
+++ b/population/client.go
@@ -429,7 +429,7 @@ func (c *Client) GetParentAreaCount(ctx context.Context, input GetParentAreaCoun
 		input.ParentAreaTypeID,
 	)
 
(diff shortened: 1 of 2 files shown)


j-s-rawat (approved):
LGTM

PR #329, opened 2022-09-22
added support for pagination

j-s-rawat:
### What
Ticket : https://trello.com/c/WjjMrc4r/5865-add-pagination-parameters-to-api-clients-go-for-getareas
Added pagination support for GetAreas via API client

### How to review
All test pass :green_circle: 

### Who can review

Anyone

### DIFF
--- a/population/client.go
+++ b/population/client.go
@@ -7,10 +7,12 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


population/client.go
```diff
@@ -36,6 +38,7 @@ type GetAreaInput struct {
 }
 
 type GetAreasInput struct {
+	Pagination       cantabular.PaginationParams
```
ChrysmOre, line 41: This shouldn't be a separate field but embedded like:
```
GetAreasInput struct{
cantabular.PaginationParams
UserAuthToken string
...
}
```

saminahbab (approved):

PR #332, opened 2022-09-28
Feature/add topic tag fields

vmilitaru:
### What

Updates the dataset response model from the dataset-api to include the `CanonicalTopic` and `SubTopics` fields.

### How to review

Read the code and run the tests.

### Who can review

Anyone

### DIFF
--- a/dataset/data.go
+++ b/dataset/data.go
@@ -32,8 +32,8 @@ type DatasetDetails struct {
 	UnitOfMeasure     string            `json:"unit_of_measure,omitempty"`
 	URI               string            `json:"uri,omitempty"`
 	IsBasedOn         *IsBasedOn        `json:"is_based_on,omitempty"`
-	CanonicalTopic 	  *TopicTagObject   `json:"canonical_topic,omitempty"`
-	SubTopics 	      *[]TopicTagObject `json:"sub_topics,omitempty"`
+	CanonicalTopic    *TopicTagObject   `json:"canonical_topic,omitempty"`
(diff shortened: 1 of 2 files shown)


dataset/data.go
```diff
@@ -496,7 +503,12 @@ func (m Metadata) ToString() string {
 	if m.RelatedDatasets != nil {
 		b.WriteString(fmt.Sprintf("Related Links: %s\n", *m.RelatedDatasets))
 	}
-
+	if m.CanonicalTopic != nil{
+		b.WriteString(fmt.Sprintf("Canonical Topic: %s\n", *m.CanonicalTopic))
+	}
+	if m.SubTopics != nil {
```
rafahop, line 509: Should we check the length of the array here too? ie. if empty, don't display it? 

Does the `Subtopics` need to be a pointer anyway?

dataset/data.go
```diff
@@ -197,6 +199,11 @@ type Metadata struct {
 	DatasetLinks Links `json:"dataset_links,omitempty"`
 }
 
+type TopicTagObject struct {
```
rafahop, line 202: Maybe we could find a better name for this type. Is there any benefit in having `Object` in it? How about simplifying it to `Topic`?

vmilitaru: `Topic` it is then :white_check_mark:

dataset/data.go
vmilitaru: Not really, just following the pattern, but I've changed this now :+1:

rafahop (approved):

PR #341, opened 2022-10-21
adds related content field to DatasetDetails response dataset model

patrickjfl:
### What

Adds support for related content field in dataset model, for use in Cantabular metadata journey.

### How to review

Read the code and run the tests.

### Who can review

Anyone

### DIFF
--- a/dataset/data_test.go
+++ b/dataset/data_test.go
@@ -92,6 +92,13 @@ func setupMetadata() Metadata {
 				"secondaryTopic2ID",
 			},
 			Survey: "census",
+			RelatedContent: &[]GeneralDetails{
+				{
+					Description: "related content description",
+					HRef:        "related content url",
+					Title:       "related content title",
+				},
+			},
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


dataset/data_test.go
```diff
@@ -139,7 +146,8 @@ func expectedData(isEmpty bool) string {
 		"Related Links: [{related dataset url related dataset title}]\n" +
 		"Canonical Topic: canonicalTopicID\n" +
 		"Subtopics: [secondaryTopic1ID secondaryTopic2ID]\n" +
-		"Survey: census\n"
+		"Survey: census\n" +
+		"Related Content: [{related content description related content url related content title}]\n"
```
cookel2, line 150: I know that this is not in these changes but the writeToFile function, below this line, should probably be deleted as it is not being used.

cookel2 (commented):
Just one comment but the changes look good to me.

dataset/data_test.go
patrickjfl: true although from reading the comment above it it looks like it might be useful in debugging?

cookel2: Yes, I did notice the comment, but I'm of the view that code should not be left in for debugging purposes alone. Debugging comments should be removed when done with. And a function like that should be stored somewhere else, such as in documentation, if it's useful to keep.

cookel2 (approved):
looks good to me

vmilitaru (approved):

PR #348, opened 2022-10-31
Add checker for metadata service

rafahop:
### What

Add health checker function for `dp-cantabular-metadata-service`

Note there is no health endpoint in there so we are making a call to the `graphql` endpoint.
Also note that we use the ext-api host as that is how it is configured to be used despite it actually using the metadata service URL. This needs to be revisited and refactored.

### How to review

Check changes make sense

### Who can review

Anyone


franmoore05 (approved):
LGTM

PR #350, opened 2022-11-03
Fix - remove call to Zebedee for data set descriptions for the new CMD or Cantabular datasets

PatrickConnollyONS:
### What

Do not ask Zebedee for data set descriptions for the new CMD or Cantabular datasets (it doesn't have their descriptions)

### How to review

Review changes and ensure tests are passing

### Who can review

Anyone

### DIFF
--- a/zebedee/client_test.go
+++ b/zebedee/client_test.go
@@ -105,7 +105,7 @@ func contentData(w http.ResponseWriter, req *http.Request) {
 	case "/bulletin/latest":
 		w.Write([]byte(`{"title":"latest release","edition":"2021","uri":"/bulletin/collection/2021"}`))
 	case "/release":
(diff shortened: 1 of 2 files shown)


zebedee/client_test.go
```diff
@@ -748,11 +740,7 @@ func TestUnitClient(t *testing.T) {
 					So(r.RelatedDatasets[0].Summary, ShouldEqual, "This is the page summary")
 					So(len(r.RelatedAPIDatasets), ShouldEqual, 2)
 					So(r.RelatedAPIDatasets[0].URI, ShouldEqual, "cantabularPageDescription")
-					So(r.RelatedAPIDatasets[0].Title, ShouldEqual, "Census data for London - population by age: 2021")
```
cookel2, line 751: We should be able to set the title in line 108 and then test it here.

zebedee/client_test.go
```diff
@@ -748,11 +740,7 @@ func TestUnitClient(t *testing.T) {
...
-					So(r.RelatedAPIDatasets[0].Summary, ShouldEqual, "Population by age in the boroughs of London")
 					So(r.RelatedAPIDatasets[1].URI, ShouldEqual, "cmdPageDescription")
-					So(r.RelatedAPIDatasets[1].Title, ShouldEqual, "A CMD dataset Title: singular")
```
cookel2, line 754: We should also be able to set the title for the CMD dataset in line 108 E.g. "relatedAPIDatasets":[{"title": "cantabularDatasetTitle","uri":"cantabularPageDescription"},{"title": "cmdDatasetTitle","uri":"cmdPageDescription"}],

cookel2 (changes requested):
Hi Patrick, I think we can add a couple of lines to check that the title is coming through from zebedee too now.

cookel2:
I also think it would be good to change the test URIs to be something like "cantabularDatasetURI" and "cmdDatasetURI" instead of "cantabularPageDescription" and "cmdPageDescription" instead now.

PR #352, opened 2022-11-10
Fix for total count

j-s-rawat:
### What

If a total count is included in the response for a Cantabular query, it always return the count of super-set and not the subset based on search criteria.This change is to return the total count based on the result of search criteria 

### How to review

All test are :green_circle: 

### Who can review

Anyone

### DIFF
--- a/cantabular/queries.go
+++ b/cantabular/queries.go
@@ -247,6 +247,33 @@ query ($dataset: String!, $text: String!, $category: String!, $limit: Int!, $off
   }
 `
 
+// QueryAreasWithoutPagination is the graphQL query to search for areas and area types which match a specific string.
+const QueryAreasWithoutPagination = `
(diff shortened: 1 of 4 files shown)


saminahbab (dismissed):

saminahbab (dismissed):

andre-urbani (approved):
LGTM

cantabular/queries.go
```diff
@@ -247,6 +245,33 @@ query ($dataset: String!, $text: String!, $category: String!, $limit: Int!, $off
   }
 `
 
+// QueryAreasWithoutPagination is the graphQL query to search for areas and area types which match a specific string.
+const QueryAreasWithoutPagination = `
```
saminahbab, line 249: a lot of the changes above are about pagination so slightly confused about the existence of this query

andre-urbani (approved):

PR #353, opened 2022-11-17
update Values to float32 type

andre-urbani:
### What

update `Values` to `float32` type, instead of `int`
Trello - https://trello.com/c/YzBzz2LY/5945-fix-json-endpoint-so-that-it-works-for-aggregate-datasets

### How to review

### Who can review

Anyone

### DIFF
--- a/cantabular/table.go
+++ b/cantabular/table.go
@@ -51,7 +51,7 @@ func (c *Client) ParseTable(table Table) (*bufio.Reader, error) {
 
 	// Obtain the CSV rows according to the cantabular dimensions and counts
 	for i, count := range table.Values {
-		row := createCSVRow(table.Dimensions, i, int(count))
+		row := createCSVRow(table.Dimensions, i, count)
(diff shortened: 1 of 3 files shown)


cantabular/table.go
```diff
@@ -51,7 +51,7 @@ func (c *Client) ParseTable(table Table) (*bufio.Reader, error) {
 
 	// Obtain the CSV rows according to the cantabular dimensions and counts
 	for i, count := range table.Values {
-		row := createCSVRow(table.Dimensions, i, count)
+		row := createCSVRow(table.Dimensions, i, int(count))
```
ChrysmOre, line 54: Kind of seems pointless to handle decimals if we are going to convert to int for the CSV?

andre-urbani: Yep. Not sure how I missed that

andre-urbani:
Required changes are now in a new PR - https://github.com/ONSdigital/dp-api-clients-go/pull/357

PR #354, opened 2022-11-22
add description field

andre-urbani:
### What

Update dimensions query to include description field 
Trello - https://trello.com/c/LryLDeBf/5940-add-meta-data-to-variable-description

### How to review

### Who can review

Anyone

### DIFF
--- a/cantabular/dimensions_test.go
+++ b/cantabular/dimensions_test.go
@@ -212,6 +212,47 @@ func TestGetGeographyDimensionsUnhappy(t *testing.T) {
 	})
 }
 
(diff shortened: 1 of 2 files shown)


ChrysmOre (commented):
LGTM but is there a unit test we could update to represent the inclusion of this field being expected?

andre-urbani:
> LGTM but is there a unit test we could update to represent the inclusion of this field being expected? Also needs rebase

It looks like there weren't actually any tests for the `GetDimensions function`, so I've added them in

PR #355, opened 2022-11-23
Fix/remove slash prefix

mr-nick17:
### What

Removed slash prefix from the `urlPath` as this was causing the `v1` from being dropped from the URL

### How to review

Sense check
Tests pass

### Who can review

!me


ChrysmOre (approved):
LGTM, how did this get through until now???

mr-nick17:
> LGTM, how did this get through until now???

Probably as I've only just started doing the frontend work 🤷‍♂️

PR #357, opened 2022-11-23
remove parse table function

andre-urbani:
### What

- Change `Values` to float32 type
- Remove `ParseTable` and related functions (these are no longer used)
Trello - https://trello.com/c/YzBzz2LY/5945-fix-json-endpoint-so-that-it-works-for-aggregate-datasets

### How to review

### Who can review

Anyone


j-s-rawat (approved):
LGTM

PR #363, opened 2022-12-16
Fix index out of bounds issue with deleting slice elements

jessjenkins:
### What

Slice element deletion causing panic when deleting due to modification of slice we're iterating over. Replaced with a reverse loop that doesn't have this problem. 

### How to review

Ensure tests pass and code is sensible.

### Who can review

Anyone but me


franmoore05 (approved):
LGTM

PR #369, opened 2023-01-10
patch: default categorisation on filter dimension

saminahbab:
### What
Add default categorisation to add dimension response

### DIFF
--- a/filter/data.go
+++ b/filter/data.go
@@ -22,13 +22,14 @@ type Dimensions struct {
 
 // Dimension represents a dimension response from the filter api
 type Dimension struct {
-	Name           string   `json:"name"`
-	ID             string   `json:"id,omitempty"`
-	Label          string   `json:"label,omitempty"`
-	URI            string   `json:"dimension_url"`
-	IsAreaType     *bool    `json:"is_area_type,omitempty"`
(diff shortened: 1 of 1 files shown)


filter/data.go
```diff
@@ -22,13 +22,14 @@ type Dimensions struct {
...
-	IsAreaType     *bool    `json:"is_area_type,omitempty"`
-	Options        []string `json:"options,omitempty"`
-	FilterByParent string   `json:"filter_by_parent,omitempty"`
+	Name                  string   `json:"name"`
+	ID                    string   `json:"id,omitempty"`
+	Label                 string   `json:"label,omitempty"`
+	URI                   string   `json:"dimension_url"`
+	IsAreaType            *bool    `json:"is_area_type,omitempty"`
+	Options               []string `json:"options,omitempty"`
+	DefaultCategorisation string   `json:"default_categorisation"`
```
ChrysmOre, line 31: Do we want to omitempty?

saminahbab: No because then the frontend does not have a clear signal for whether there is a default categorisation, we thought last month that an empty string can show that there is no default categorisation set

ChrysmOre (approved):

PR #371, opened 2023-01-11
Add UsageNotes to datasetDetails

DavidSubiros:
### What

Add UsageNotes back to DatasetDetails, to prevent observation-api from breaking when it's updated.

### How to review

- Make sure change is ok and compatible with observation api.

### Who can review

@redhug1


redhug1 (approved):
LGTM

PR #375, opened 2023-01-12
Added support for dimensions description

j-s-rawat:
### What
Implemented a api client function to return descriptions of passed dimensions

### How to review
All tests are :green_circle: 

### Who can review

Anyone

### DIFF
--- a/cantabular/dimensions.go
+++ b/cantabular/dimensions.go
@@ -104,6 +104,33 @@ func (c *Client) GetDimensions(ctx context.Context, req GetDimensionsRequest) (*
 	return &resp.Data, nil
 }
 
+// GetDimensionsDescription performs a graphQL query to get the description of the passed dimensions
+func (c *Client) GetDimensionsDescription(ctx context.Context, req GetDimensionsDescriptionRequest) (*GetDimensionsResponse, error) {
+	resp := &struct {
+		Data   GetDimensionsResponse `json:"data"`
+		Errors []gql.Error           `json:"errors,omitempty"`
+	}{}
+
+	data := QueryData{
+		Dataset:   req.Dataset,
+		Variables: req.DimensionNames,
+	}
+
+	if err := c.queryUnmarshal(ctx, QueryDimensionsDescription, data, resp); err != nil {
+		return nil, err
+	}
+
+	if resp != nil && len(resp.Errors) != 0 {
+		return nil, dperrors.New(
+			errors.New("error(s) returned by graphQL query"),
+			resp.Errors[0].StatusCode(),
+			log.Data{"errors": resp.Errors},
+		)
+	}
+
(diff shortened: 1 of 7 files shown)


cantabular/dimensions.go
```diff
@@ -104,6 +104,34 @@ func (c *Client) GetDimensions(ctx context.Context, req GetDimensionsRequest) (*
 	return &resp.Data, nil
 }
 
+// GetDimensions performs a graphQL query to obtain all the non-geography dimensions for the provided
```
ChrysmOre, line 107: This comment is wrong

cantabular/dimensions_test.go
```diff
@@ -1999,6 +2059,36 @@ var mockRespBodySearchDimensions = `{
 	}
 }
 `
+var mockRespBodyGetDimensionsDescription = `{
```
ChrysmOre, line 2062: There's no Description tested here?

population/dimensions.go
```diff
@@ -124,6 +130,62 @@ func (c *Client) GetDimensions(ctx context.Context, input GetDimensionsInput) (G
 	return dimensions, nil
 }
 
+func (c *Client) GetDimensionsDescription(ctx context.Context, input GetDimensionsDescriptionInput) (GetDimensionsResponse, error) {
+	logData := log.Data{
+		"method":       http.MethodGet,
+		"dimensionIDs": input.DimensionIDs,
+	}
+
+	urlPath := fmt.Sprintf("population-types/%s/dimensions/description", input.PopulationType)
```
ChrysmOre, line 139: what if there's a dimension called 'description' in population-types-api?

ChrysmOre (commented):
Few comments also:
- Does this endpoint actually exist? How can you be sure this works if there's no endpoint to test with?
  **=>** _I have tested this using "hello" endpoint of dimension API. Also there is an endpoint that is introduced in the dp-population-types-api but as that is dependent on ap-client I will post that PR later._ 

- The endpoint in question /dimensions/description probably isn't the best way to do it.. what if there's a dimension called 'description'? Would probably be better to use dimensions-description rather than dimensions/description
**=>** _If there is a dimension called description and we are searching for that dimension, then the URL will be 
`/dimensions?q=description` 
If we are looking for the description of that dimension called 'description' 
`/dimensions/description?q=description`_  
**I will update it to dimensions-description as that will be inline with existing endpoints**

- Also what is the difference between this and regular GetDimensions? Is it the same with added Description field? Any reason we couldn't just add Description to GetDimensions?
**=>** _GetDimenstions uses a query where we user can do a fuzzy search on the dimension name passed to it. But for description we need an exact match on the passed description/dimensions._ 
 
Lots of questions sorry :)

ChrysmOre (approved):

PR #381, opened 2023-01-31
feat: GetDimensionCategories client method addded

saminahbab:
Client method for population types api added

### DIFF
--- a/population/dimensions.go
+++ b/population/dimensions.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 
 	"github.com/ONSdigital/dp-api-clients-go/v2/clientlog"
 	dperrors "github.com/ONSdigital/dp-api-clients-go/v2/errors"
@@ -24,10 +25,28 @@ type Dimension struct {
 	QualityStatementText string     `json:"quality_statement_text"`
 }
 
+type DimensionCategory struct {
+	Id         string                  `json:"id"`
+	Label      string                  `json:"label"`
+	Categories []DimensionCategoryItem `json:"categories"`
+}
+
+type DimensionCategoryItem struct {
+	ID    string `json:"id"`
+	Label string `json:"label"`
+}
+
 type Category struct {
 	ID                   string `json:"id"`
 	Label                string `json:"label"`
 	QualityStatementText string `json:"quality_statement_text"`
+}
+
+type GetDimensionCategoryInput struct {
+	AuthTokens
+	PaginationParams
+	PopulationType string
+	Dimensions     []string
 }
 
 type GetDimensionsInput struct {
@@ -48,6 +67,11 @@ type GetCategorisationsInput struct {
 	PaginationParams
 	PopulationType string
 	Dimension      string
+}
+
+type GetDimensionCategoriesResponse struct {
+	PaginationResponse
+	Categories []DimensionCategory `json:"items"`
 }
 
 // GetDimensionsResponse is the response object for GetDimensions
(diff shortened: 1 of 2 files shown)


population/dimensions.go
```diff
@@ -24,12 +24,31 @@ type Dimension struct {
 	QualityStatementText string     `json:"quality_statement_text"`
 }
 
+// Slightly confusing given very similar to above, but felt no other way. Remove this comment when others approve naming
+type DimensionCategory struct {
```
ChrysmOre, line 28: Seems ok to me

population/dimensions.go
```diff
@@ -24,12 +24,31 @@ type Dimension struct {
...
 	ID                   string `json:"id"`
 	Label                string `json:"label"`
 	QualityStatementText string `json:"quality_statement_text"`
 }
 
+type GetDimensionCategoryInput struct {
+	AuthTokens
+	PaginationParams
+	PopulationType string
+	Dimensions     string
```
ChrysmOre, line 49: DImensions? Is this supposed to be singular or a slice?

saminahbab: the swagger listed it as a comma separated url query param so i left it as a string here

ChrysmOre: Hmm I think it would be better to have it as a slice for the api client? And convert it to the appropriate string format when creating the URL param. What do we do elsewhere?

population/dimensions_test.go
```diff
@@ -15,6 +15,87 @@ import (
...
+			}
+			res, err := client.GetDimensionCategories(context.Background(), input)
+
+			Convey("it should return a list of dimensions", func() {
+				So(err, ShouldBeNil)
+				So(res, ShouldResemble, categories)
+			})
+
+		})
+
```
ChrysmOre, line 95: Lots of extra whitespace here

population/dimensions_test.go
```diff
@@ -15,6 +15,87 @@ import (
...
+		input := GetDimensionCategoryInput{
+			AuthTokens: AuthTokens{
+				UserAuthToken:    userAuthToken,
+				ServiceAuthToken: serviceAuthToken,
+			},
+			PopulationType: populationType,
+			Dimensions:     []string{dimension},
+		}
+
+		client.GetDimensionCategories(context.Background(), input)
```
ChrysmOre, line 37: I think the calls to the func being tested should be in their own convey statement or it be mentioned in the one it's in

population/dimensions_test.go
```diff
@@ -15,6 +15,87 @@ import (
...
+		stubClient := newStubClient(&http.Response{Body: ioutil.NopCloser(bytes.NewReader(nil))}, nil)
+		client, err := NewWithHealthClient(health.NewClientWithClienter("", "http://test.test:2000/v1", stubClient))
+		So(err, ShouldBeNil)
+		input := GetDimensionCategoryInput{
+			AuthTokens: AuthTokens{
+				UserAuthToken:    userAuthToken,
+				ServiceAuthToken: serviceAuthToken,
+			},
+			PopulationType: populationType,
+			Dimensions:     []string{dimension},
```
ChrysmOre, line 34: Probably best to test with more than one dimension to be sure the comma separation is good

population/dimensions_test.go
```diff
@@ -15,6 +15,87 @@ import (
...
+		}
+
+		client.GetDimensionCategories(context.Background(), input)
+		Convey("It should call the dimensions category endpoint", func() {
+			calls := stubClient.DoCalls()
+			So(calls, ShouldNotBeEmpty)
+			fmt.Println(calls[0].Req.URL.String())
```
ChrysmOre, line 41: debug

population/dimensions_test.go
saminahbab: added a test with 2 dimensions

population/dimensions_test.go
saminahbab: error calls and 404 call tests added checking for error

population/dimensions_test.go
saminahbab: this has been added

population/dimensions.go
```diff
@@ -72,6 +96,65 @@ type GetBaseVariableResponse struct {
 	Label string `json:"label"`
 }
 
+func (c *Client) GetDimensionCategories(ctx context.Context, input GetDimensionCategoryInput) (GetDimensionCategoriesResponse, error) {
```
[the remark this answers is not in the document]
saminahbab: the style of the function signature is actually the same as the other functions in the file and client struct, so I would say it is fine in this instance, other packages within the api clients go may be different, i accept

j-s-rawat (approved):
LGTM

PR #383, opened 2023-02-07
add custom flag to filter model

ChrysmOre:
### What

Added `custom` flag to filter model

Includes some mild tidy up

### How to review

Check changes make sense

### Who can review

Anyone

### DIFF
--- a/cantabular/client.go
+++ b/cantabular/client.go
@@ -178,7 +178,7 @@ func (c *Client) errorResponse(url string, res *http.Response) error {
 	}
 
 	if len(b) == 0 {
-		b = []byte("[response body empty]")
+		b = []byte(`{"message": "(error response body empty)"}`)
 	}
 
 	var resp ErrorResponse
--- a/filter/contract.go
+++ b/filter/contract.go
@@ -1,4 +1,5 @@
 package filter
+
 // Find in this file up to date request/responses to some of the dp-filter-api
 // endpoints. For some reason the existing functions are largely out of date
 // with how the API actually behaves, but are going to create fresh functions
@@ -8,7 +9,7 @@ package filter
 
 // AuthHeaders represents the common set of headers required for making
 // authorized requests
-type AuthHeaders struct{
+type AuthHeaders struct {
 	UserAuthToken    string
 	ServiceAuthToken string
 	CollectionID     string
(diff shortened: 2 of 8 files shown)


filter/contract.go
```diff
@@ -8,35 +9,55 @@ package filter
...
 	Self       Link `json:"self,omitempty"`
 	Version    Link `json:"version,omitempty"`
 }
+
+type CreateFlexBlueprintRequest struct {
+	Dataset        Dataset          `json:"dataset"`
+	Dimensions     []ModelDimension `json:"dimensions"`
+	PopulationType string           `json:"population_type"`
+	Custom         bool             `json:"custom"`
+	CollectionID   string           `json:"collectio"`
```
cookel2, line 51: This looks like a typo but I'm not sure whether it should be this or "collection" or "collecton_id"?
```suggestion
	CollectionID   string           `json:"collection"`
```

cantabular/client.go
```diff
@@ -178,7 +178,7 @@ func (c *Client) errorResponse(url string, res *http.Response) error {
 	}
 
 	if len(b) == 0 {
-		b = []byte("[response body empty]")
+		b = []byte(`{"message": "(error response body empty)"}`)
```
cookel2, line 181: To make this message clear that it's referring to an error response from cantabular it would be good to say something like this:
```suggestion
		b = []byte(`{"message": "(error response body empty)"}`)
```
```suggestion
		b = []byte(`{"message": "(cantabular error has empty response body)"}`)
```

filter/contract.go
```diff
@@ -8,35 +9,55 @@ package filter
...
 }
 
 // FilterLinks represents the links object for /filters related endpoints
-type FilterLinks struct{
+type FilterLinks struct {
 	Dimensions Link `json:"dimensions,omitempty"`
 	Self       Link `json:"self,omitempty"`
 	Version    Link `json:"version,omitempty"`
 }
+
```
cookel2, line 45: It would be good to add doc for the CreateFlexBlueprintRequest and createFlexBlueprintResponse structs

cookel2 (commented):
Just a couple of suggestions Josh, thanks.

cantabular/client.go
ChrysmOre: The way we wrap errors at each level this won't be necessary. It will always look like `failed to do a: failed to post query: {error response body empty}. We also don't want to mention cantabular in any responses we return to the user

cookel2: That makes sense, thanks.

cookel2 (approved):
looks good to me

PR #385, opened 2023-02-07
Fix QueryCategorisations graphql

thomasridd:
### What

Shifted Quality Statement to categorisation level from individual category 

### How to review

Draft: Please pick up this PR and work through any other changes required for the model 

### Who can review

@andre-urbani

### DIFF
--- a/cantabular/queries.go
+++ b/cantabular/queries.go
@@ -444,14 +444,14 @@ query ($dataset: String!, $text: String!) {
 		    totalCount
 		    edges{
 		      node{
+				meta {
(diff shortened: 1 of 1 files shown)


ChrysmOre (commented):
Is there not a test that should be updated to reflect this change?

thomasridd:
I agree on the principle but if adding a test will take significant time this is needed to unblock sandbox use for demos. 
It shouldn't be this way but it is

andre-urbani (approved):

PR #388, opened 2023-02-10
Added GetMetadata

j-s-rawat:
### What
Ticket : https://trello.com/c/ELlEBmEn/6022-create-end-point-for-new-custom-filter-flex-api
Introduced API client function to get the default dataset metadata

### How to review

All test are :green_circle: 
### Who can review

Anyone

### DIFF
--- a/population/population_type_metadata.go
+++ b/population/population_type_metadata.go
@@ -13,19 +13,21 @@ import (
 	"github.com/pkg/errors"
 )
 
-type GetMetaDataInput struct {
+// GetPopulationTypeMetadataInput is a model with auth token and population type
+type GetPopulationTypeMetadataInput struct {
 	AuthTokens
 	PopulationType string
 }
 
-type GetMetadataResponse struct {
-	PopulationType   string `json:"population-type"`
+// GetPopulationTypeMetadataResponse model with contain the metadata for poulation type
+type GetPopulationTypeMetadataResponse struct {
+	PopulationType   string `json:"population_type"`
 	DefaultDatasetID string `json:"default_dataset_id"`
 	Edition          string `json:"edition"`
 	Version          int    `json:"version"`
 }
 
-func (c *Client) GetMetadata(ctx context.Context, input GetMetaDataInput) (GetMetadataResponse, error) {
+func (c *Client) GetPopulationTypeMetadata(ctx context.Context, input GetPopulationTypeMetadataInput) (GetPopulationTypeMetadataResponse, error) {
 	logData := log.Data{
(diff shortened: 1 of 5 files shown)


population/default_dataset_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+type GetMetadataResponse struct {
+	PopulationType   string `json:"population-type"`
+	DefaultDatasetID string `json:"default_dataset_id"`
+	Edition          string `json:"edition"`
+	Version          int    `json:"version"`
+}
+
+func (c *Client) GetDefaultDatasetMetadata(ctx context.Context, input GetMetaDataInput) (GetMetadataResponse, error) {
```
ChrysmOre, line 28: We're not getting metadata about the default dataset, we're getting metadata about the population type, which includes the id of the default dataset. This function would be bette called GetPopulationTypeMetadata, and the Inputas/Responses should be the same

population/population_type_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+	"net/http"
+	"net/url"
+
+	"github.com/ONSdigital/dp-api-clients-go/v2/clientlog"
+	dperrors "github.com/ONSdigital/dp-api-clients-go/v2/errors"
+	"github.com/ONSdigital/log.go/v2/log"
+	"github.com/pkg/errors"
+)
+
+type GetPopulationTypeMetadataInput struct {
```
jessjenkins, line 16: All exported structs and functions should have GoDoc comments explaining their purpose. See [other packages in this repo](https://github.com/ONSdigital/dp-api-clients-go/blob/main/image/image.go) for examples.

population/population_type_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+	vals := url.Values{}
+	req, err := c.createGetRequest(ctx, input.UserAuthToken, input.ServiceAuthToken, uri, vals)
+	if err != nil {
+		return GetPopulationTypeMetadataResponse{}, dperrors.New(
+			err,
+			dperrors.StatusCode(err),
+			logData,
+		)
+	}
+	clientlog.Do(ctx, "getting metadata", service, req.URL.String(), logData)
```
jessjenkins, line 42: Although fine in this context, when this library is used in an app "getting metadata" in the log output might become ambiguous, better to be more specific.
```suggestion
	clientlog.Do(ctx, "getting population type metadata", service, req.URL.String(), logData)
```

andre-urbani (approved):
LGTM

PR #393, opened 2023-02-21
handle cantabular table level error

j-s-rawat:
### What

Passing  the table level error from cantabular to FE
### How to review

All test are :green_circle: 

 Anyone

### DIFF
--- a/cantabular/contract.go
+++ b/cantabular/contract.go
@@ -225,9 +225,10 @@ type GetParentAreaCountResult struct {
 }
 
 type GetBlockedAreaCountResult struct {
-	Passed  int `json:"passed"`
-	Blocked int `json:"blocked"`
-	Total   int `json:"total"`
+	Passed         int     `json:"passed"`
+	Blocked        int     `json:"blocked"`
+	Total          int     `json:"total"`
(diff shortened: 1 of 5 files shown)


cantabular/contract.go
```diff
@@ -225,9 +225,10 @@ type GetParentAreaCountResult struct {
 }
 
 type GetBlockedAreaCountResult struct {
-	Passed  int `json:"passed"`
-	Blocked int `json:"blocked"`
-	Total   int `json:"total"`
+	Passed         int     `json:"passed"`
+	Blocked        int     `json:"blocked"`
+	Total          int     `json:"total"`
+	TableLeveError *string `json:"table_level_error,omitempty"`
```
ChrysmOre, line 231: I don't think this needs to be a pointer? Omitempty should work as expected with an empty string. Also `TableError` might be enough rather than `TableLevelError`, trying to keep names as concise as we can

j-s-rawat: Will check with FE developer if they will prefer the field to be missing or have null in it and change accordingly

andre-urbani (approved):
LGTM

PR #411, opened 2023-03-23
New function to read version etag 

rafahop:
### What

- Add new function returning headers for Get version. Currently only containing etag
- Add unit test for the happy path of GetVersion and for the new function GetVersionWithHeaders
- Refactor checks of the response headers in unit test so to expand it to collection id, florence token and download service token.

### How to review

Check the code and test look ok

See https://github.com/ONSdigital/dp-dataset-api/pull/416 for the change in dataset api introducing the etag in the header

### Who can review

Anyone.

### DIFF
--- a/dataset/dataset_test.go
+++ b/dataset/dataset_test.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


dataset/dataset_test.go
```diff
@@ -221,11 +236,98 @@ func TestClient_HealthChecker(t *testing.T) {
...
+					ServiceToken:         serviceAuthToken,
+					CollectionId:         collectionID,
+					DownloadServiceToken: downloadServiceAuthToken,
+				}
+				checkRequestBase(httpClient, http.MethodGet, expectedUrl, expectedHeaders)
+			})
+		})
+
+		Convey("when GetVersionWithHeaders is called", func() {
+			got, h, err := datasetClient.GetVersionWithHeaders(ctx, userAuthToken, serviceAuthToken, downloadServiceAuthToken, collectionID, datasetId, edition, versionString)
```
s-xton, line 296: It is your intension that the THEN clauses are tested in isolation with multiple calls to GetVersionWithHeaders ?

s-xton (approved):

s-xton (changes requested):
see code comment

dataset/dataset_test.go
rafahop: Not really, but thought it'd read better using separate conveys but I'll join them in one to improve performance 👍

s-xton (approved):
can you raise a story to move the if statement on the service token to another function pls, not needed now but nice to clean up later

PR #412, opened 2023-03-24
cantabular - handle errors from dataset.table

ChrysmOre:
### What

Handle errors from dataset.Table. Either return full string or lookup from table depending.

### How to review

Check changes make sense

### Who can review

Anyonne


franmoore05 (approved):
LGTM

PR #415, opened 2023-03-30
Increase default limit on GetPopulationTypes response

thomasridd:
### What

Increase default limit on GetPopulationTypes response

### How to review

Sense check

### Who can review

!me

### DIFF
--- a/population/population_types.go
+++ b/population/population_types.go
@@ -3,6 +3,7 @@ package population
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
 
@@ -19,6 +20,7 @@ type PopulationType struct {
 }
 
 type GetPopulationTypesInput struct {
+	PaginationParams
 	DefaultDatasets bool
 	AuthTokens
 }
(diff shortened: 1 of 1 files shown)


population/population_types.go
```diff
@@ -42,7 +44,12 @@ func (c *Client) GetPopulationTypes(ctx context.Context, input GetPopulationType
 		"method": http.MethodGet,
 	}
 
-	urlPath := "population-types"
+	limit := 1000
```
j-s-rawat, line 47: Do we want it to be an enormously high no. for future proofing

thomasridd: I've defaulted to passing no parameters but I suspect we are now having to pass in the number 1000 in every place we use the method

thomasridd: or perhaps 10,000 as you suggest - though I think if we get over 100 our servers will burst into fire

population/population_types.go
```diff
@@ -43,6 +45,12 @@ func (c *Client) GetPopulationTypes(ctx context.Context, input GetPopulationType
 	}
 
 	urlPath := "population-types"
+	if input.Limit > 0 {
```
j-s-rawat, line 48: don't think we need these tests, 'limit' and 'offset' will be initialised to 0 when the structure is passed, and currently if 0 is passed as limit code defaults it to '20' So simply adding what ever limit and offset that is passed in pagination should be ok

j-s-rawat (approved):
LGTM

PR #429, opened 2024-01-22
Feature/changes to improve census observations

franmoore05:
### What

Added some new functions to be used as part of the stabilisation work for dp-population-types-api.  Some of these appear to be duplicates of existing functionality but there are differences in what they return.  These were added to avoid breaking changes.

### How to review

Ensure the tests pass and the code makes sense.

### Who can review

Anyone ONS.


cantabular/queries.go
```diff
@@ -600,6 +600,16 @@ query ($dataset: String!, $variables: [String!]! ) {
...
 
+const QueryValues = `
+query($dataset: String!, $variables: [String!]!, $filters: [Filter!]) {
+	dataset(name: $dataset) {
+		table(variables: $variables, filters: $filters) {
+			values
+            error
+		}
+	}
+}`
```
lindenmckenzie, line 611: Indentation doesn't look right to me here

cantabular/table.go
```diff
@@ -35,6 +37,37 @@ type Table struct {
...
+type ObservationDimension struct {
+	Dimension   string `bson:"dimension"           json:"dimension"`
+	DimensionID string `bson:"dimension_id"           json:"dimension_id"`
+	Option      string `bson:"option"           json:"option"`
+	OptionID    string `bson:"option_id"           json:"option_id"`
+}
+
+type GetObservationResponse struct {
+	Dimensions  []ObservationDimension `bson:"dimensions"           json:"dimensions"`
+	Observation float32                `bson:"observation"   json:"observation"`
```
lindenmckenzie, line 49: Indentation could be tidied up here. Additionally - do we need the `bson`? There seems to be a mixture of approaches here.

cantabular/table.go
```diff
@@ -96,6 +129,60 @@ func GraphQLJSONToCSV(ctx context.Context, r io.Reader, w io.Writer) (int32, err
 	return rowCount, nil
 }
 
+// GraphQLJSONToJson converts a JSON response in r to a different schema on w, returning the response
+// if an error happens, the process is aborted and the error is returned.
+func GraphQLJSONToJson(ctx context.Context, r io.Reader, w io.Writer) (GetObservationsResponse, error) {
```
lindenmckenzie, line 134: Is there a unit test for this?

cantabular/table.go
```diff
@@ -96,6 +129,60 @@ func GraphQLJSONToCSV(ctx context.Context, r io.Reader, w io.Writer) (int32, err
...
+	// decode 'data' and 'error' fields
+	for dec.More() {
+		field, err := dec.DecodeName()
+		if err != nil {
+			return GetObservationsResponse{}, fmt.Errorf("error decoding field: %w", err)
+		}
+		switch field {
+		case "data":
+			if getObservationsResponse, err = decodeDataFieldsJson(ctx, dec, w); err != nil {
+				// null values for 'dataset' or 'table' are ok as long as the error is reported under the 'errors' field
```
lindenmckenzie, line 159: I'm not entirely clear on how this functionality is working - It seems to be that if it returns an error here, the loop stops being executed and then we go onto the `EndComposite` and checking `errData` !=nil where it will error. Maybe I'm missing something? I can see it's been implemented in this way in other places in this file.

cantabular/table.go
```diff
@@ -275,6 +482,32 @@ func decodeValues(ctx context.Context, dec jsonstream.Decoder, dims Dimensions,
...
+
+	// Obtain the Json objects according to the cantabular dimensions and counts
+	for ti := dims.NewIterator(ctx); dec.More(); {
+		count, err := dec.DecodeNumber()
+		if err != nil {
+			return GetObservationsResponse{}, fmt.Errorf("error decoding count: %w", err)
+		}
+		row, err := ti.createJsonObject(dims, count.String())
+		if err != nil {
+			return GetObservationsResponse{}, fmt.Errorf("error parsing a csv row: %w", err)
```
lindenmckenzie, line 498: Think this error message should be a json node not csv row?

cantabular/table.go
```diff
@@ -275,6 +482,32 @@ func decodeValues(ctx context.Context, dec jsonstream.Decoder, dims Dimensions,
...
+		}
+		obsResp.Encode(row)
+		getObservationsResponse.Observations = append(getObservationsResponse.Observations, row)
+
+		if err := ti.Next(); err != nil {
+			return GetObservationsResponse{}, fmt.Errorf("error iterating to next row: %w", err)
```
lindenmckenzie, line 504: Think this error should be json node not row?

cantabular/table.go
```diff
@@ -309,3 +542,33 @@ func (it *Iterator) createCSVRow(dims []Dimension, count string) ([]string, erro
...
+		obsDim.Option = category.Label
+		obsDim.OptionID = category.Code
+		obsDimArr = append(obsDimArr, obsDim)
+
+	}
+
+	obsResponse.Dimensions = obsDimArr
+	observation, err := strconv.ParseFloat(count, 32)
+	if err != nil {
+		return GetObservationResponse{}, fmt.Errorf("invalid observation %d, : %w", 0, err)
```
lindenmckenzie, line 570: Why are you passing 0 here for the digit?

cantabular/static_dataset.go
```diff
@@ -153,3 +153,91 @@ func (c *Client) StaticDatasetQueryStreamCSV(ctx context.Context, req StaticData
 	// Stream is responsible for closing the response body
 	return rowCount, stream.Stream(ctx, res.Body, transform, consume)
 }
+
+// Checks the number of observations returned from a cantabular query
+func (c *Client) CheckQueryCount(ctx context.Context, req StaticDatasetQueryRequest) (int, error) {
```
lindenmckenzie, line 158: Is there a unit test for this?

cantabular/static_dataset.go
```diff
@@ -153,3 +153,91 @@ func (c *Client) StaticDatasetQueryStreamCSV(ctx context.Context, req StaticData
...
+
+	return len(q.Data.Dataset.Table.Values), nil
+}
+
+// StaticDatasetQueryStreamJson performs a StaticDatasetQuery call
+// and then starts 2 go-routines to transform the response body into a Json stream and
+// consume the transformed output with the provided Consumer concurrently.
+// Returns a json formatted response
+// Use this method if large query responses are expected.
+func (c *Client) StaticDatasetQueryStreamJson(ctx context.Context, req StaticDatasetQueryRequest, consume Consumer) (GetObservationsResponse, error) {
```
lindenmckenzie, line 215: Is there a unit test for this?

lindenmckenzie (commented):
Few fairly minor comments and some bits I didn't quite follow - as a general question, was there any Go packages to help us do this as translating GraphQL to JSON seems like a fairly standard process, although maybe less so for Go apps.

cantabular/queries.go
franmoore05: This has now been fixed.

cantabular/table.go
franmoore05: The bson part has now been removed, it had been copied across from previous code.

cantabular/table.go
franmoore05: This is fully tested as part of component tests in dp-population-types-api that consumes it.

cantabular/table.go
franmoore05: This has been updated

cantabular/table.go
franmoore05: This has been updated

cantabular/table.go
franmoore05: The response has now been amended, in the code it was taken from this referred to a row but this is no longer relevant.

cantabular/static_dataset.go
franmoore05: This is fully tested as part of component tests in dp-population-types-api that consumes it.

cantabular/static_dataset.go
franmoore05: This is fully tested as part of component tests in dp-population-types-api that consumes it.

cantabular/table.go
franmoore05: This is taken from code that SCC provided for us a while ago so my very basic understanding of it is that if the fields dataset or table are returned from the Cantabular graphQL query then there should be something in the errors field instead - an example response might look something like this in this case:

{
  "data": {
    "dataset": {
      "table": null
    }
  },
  "errors": [
    {
      "message": "400 Bad Request: variable at position 2 does not exist",
      "locations": [
        {
          "line": 2,
          "column": 23
        }
      ],
      "path": [
        "dataset",
        "table"
      ]
    }
  ]
}

cantabular/table.go
lindenmckenzie: Just this one to go.

franmoore05: This should now be fixed

lindenmckenzie (approved):

PR #438, opened 2024-08-30
Add zebedee GetPageData

jondewijones:
### What

Add `GetPageData` func to Zebedee client

### Who can review

Anyone

### DIFF
--- a/zebedee/client.go
+++ b/zebedee/client.go
@@ -306,6 +306,38 @@ func (c *Client) GetPageTitle(ctx context.Context, userAccessToken, collectionID
 	return pt, nil
 }
 
+// GetPageType retrieves the type of content and page title for a given URI from zebedee
+func (c *Client) GetPageType(ctx context.Context, userAccessToken, collectionID, lang, uri string) (PageType, error) {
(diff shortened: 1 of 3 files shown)


zebedee/client.go
```diff
@@ -306,6 +306,38 @@ func (c *Client) GetPageTitle(ctx context.Context, userAccessToken, collectionID
...
+	}
+
+	var pt PageTypeAndMetadata
+	if err = json.Unmarshal(b, &pt); err != nil {
+		return pt, err
+	}
+
+	return pt, nil
+}
+
```
lindenmckenzie, line 340: I feel like I"m missing something here - these look identical.

jondewijones: `GetType` returns just the type and `GetPageTypeAndMetadata` returns type and some metadata. Was using type first, then realised we needed more data so added a new func and left the old one. Happy to remove `GetType` if we think it's superfluous

jondewijones: As per our call i've remove both funcs and replaced them with a `GetPageData` func that returns a more generic page response

lindenmckenzie (approved):

PR #462, opened 2026-04-23
fix addition of authz header to zebedee client

jessjenkins:
### What

The addition of `Authorization` header to zebedee client requests was incorrectly adding an extra `Bearer ` to JWT tokens (causing `Bearer Bearer <token>`. This uses a new function in an updated version of dp-net that now adds the token unmodified from what was supplied by this library. 

However, the zebedee client itself now will add `Bearer` only where necessary so that either fully qualified user auth tokens containing the Bearer prefix or legacy service auth tokens that omit it can be used using the same `authToken` parameter in the function call. 

Give the above, any references to `userAuthToken` or `userAceessToken` have been renamed to the less restrivtive `authToken`. 

### How to review

Ensure only expected modification to zebedee client and tests pass.

### Who can review

Anyone


zebedee/client.go
```diff
@@ -194,16 +193,26 @@ func (c *Client) get(ctx context.Context, userAccessToken, path string) ([]byte,
 	return b, resp.Header, err
 }
 
+func addAuthHeadersToRequest(req *http.Request, authToken string) {
+	// Service auth tokens dont usually contain `Bearer` already so add it if it is missing
+	if !strings.HasPrefix(authToken, "Bearer ") {
+		authToken = "Bearer " + authToken
+	}
+	// Note: Florence header ("X-Florence-Token") will be depricated in future and only the Authorization header will
```
lindenmckenzie:
```suggestion
	// Note: Florence header ("X-Florence-Token") will be deprecated in future and only the Authorization header will
```

zebedee/client.go
```diff
@@ -194,16 +193,26 @@ func (c *Client) get(ctx context.Context, userAccessToken, path string) ([]byte,
 	return b, resp.Header, err
 }
 
+func addAuthHeadersToRequest(req *http.Request, authToken string) {
+	// Service auth tokens dont usually contain `Bearer` already so add it if it is missing
```
lindenmckenzie:
```suggestion
	// Service auth tokens don't usually contain `Bearer` already so add it if it is missing
```

lindenmckenzie (dismissed):

lindenmckenzie (approved):

